PR #3, opened 2020-06-03
lib/test: add initial prototype for wiby --test

andrewhughes101:
Intial protoype of tool, as demoed in the package maintenance meeting yesterday.

This PR contains the functionality to run `wiby --test dependent_url` which will take the HEAD commit from the repository the tool is being run from, and then update the package json of the dependent module provided to the tool as dependent_url.

### DIFF
--- a/bin/wiby.js
+++ b/bin/wiby.js
@@ -1,12 +1,26 @@
 #!/usr/bin/env node
 
+require('dotenv').config()
 const test = require('../lib/test')
 const result = require('../lib/result')
 
-if (process.argv[2] === 'test') {
-  test()
+const args = require('yargs')
+  .option('test', {
+    alias: 't',
+    describe: 'Test your dependents'
+  })
+  .option('result', {
+    alias: 'r',
+    describe: 'Get the result of your tests'
+  })
+  .argv
+
+if (args.test) {
+  test(args.test)
 }
 
-if (process.argv[2] === 'result') {
+if (args.result) {
   result()
 }
+
+// Usage: wiby --test URL
--- a/lib/test.js
+++ b/lib/test.js
@@ -1,3 +1,105 @@
-module.exports = () => {
-  console.log('wiby test')
+const fsPromises = require('fs').promises
+const fetch = require('node-fetch')
+const github = require('./github')
+
+module.exports = async function (url) {
+  const dep = await getLocalPackageName()
+  console.log('Got local package name: ', dep)
+  const [org, repo] = getOrgRepo(url)
+  console.log(`Got org: ${org} and repo: ${repo}`)
+  const packageJSON = await github.getPackageJson(org, repo)
+  if (!checkPackageInPackageJSON(dep, packageJSON)) {
+    throw new Error('Dependency not found in package.json')
+  }
+  const patch = await getPatch(dep)
+  console.log('Created patch: ', patch)
+  const patchedPackageJSON = applyPatch(patch, dep, packageJSON)
+  await pushPatch(patchedPackageJSON, org, repo, dep)
+}
+
+const getCommitHash = module.exports.getCommitHash =
+async function getCommitHash (owner, repo) {
+  const headSha = (await github.getShas(owner, repo))[0]
+  return headSha
+}
+
+const checkPackageInPackageJSON = module.exports.checkPackageInPackageJSON =
+function checkPackageInPackageJSON (dep, packageJSON) {
+  return Object.prototype.hasOwnProperty.call(packageJSON.dependencies, dep)
+}
+
+const getOrgRepo = module.exports.getOrgRepo =
+function getOrgRepo (url) {
+  const repoOrgArr = (url.split('github.com'))[1].split('/')
+  const org = repoOrgArr[1]
+  const repo = repoOrgArr[2]
+  return [org, repo]
+}
+
+const getPatch = module.exports.getPatch =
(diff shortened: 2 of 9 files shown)


lib/test.js
```diff
@@ -1,3 +1,107 @@
-module.exports = () => {
-  console.log('wiby test')
+const fs = require('fs')
+const fetch = require('node-fetch')
+const { promisify } = require('util')
+const github = require('./github')
+const readFileAsync = promisify(fs.readFile)
```
BethGriggs, line 5: This is a nit type comment- but how would you feel about using the fs Promise API? It may reduce the imports a little https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api

andrewhughes101: Thanks I havent seen that before, I'll push an update using the Promise api

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+  const file = JSON.stringify(packageJSON, null, 2) + '\n' // assumes package.json is using two spaces
+  const encodedFile = Buffer.from(file).toString('base64')
+  const message = `wiby: update ${dep}`
+  const branch = `wiby-${dep}`
+
+  const [headSha, treeSha] = await github.getShas(owner, repo)
+  const blobSha = await github.createBlob(owner, repo, encodedFile)
+  const newTreeSha = await github.createTree(owner, repo, treeSha, blobSha)
+  const commitSha = await github.createCommit(owner, repo, message, newTreeSha, headSha)
+  await github.createBranch(owner, repo, commitSha, branch)
```
dominykas, line 103: I haven't looked in detail, but I think it might be possible to do all of the above in a single API request: https://octokit.github.io/rest.js/v18#repos-create-or-update-file-contents.

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+  }
+  packageJSON.dependencies[dep] = patch
+  return packageJSON
+}
+
+async function pushPatch (packageJSON, owner, repo, dep) {
+  const file = JSON.stringify(packageJSON, null, 2) + '\n' // assumes package.json is using two spaces
+  const encodedFile = Buffer.from(file).toString('base64')
+  const message = `wiby: update ${dep}`
+  const branch = `wiby-${dep}`
```
dominykas, line 97: Does `dep` need sanitization here? Is `@` an allowed character in a git branch (in case `dep` is a scoped package).

Should the `branch` and `message` also include some ref back to the hash / branch / PR of the dep being tested?

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+  if (!org && !repo) {
+    org = 'undefined'
+    repo = 'undefined'
+  }
+  return [org, repo]
+}
+
+const fetchRegistryInfo = module.exports.fetchRegistryInfo =
+async function fetchRegistryInfo (dep) {
+  const resp = await fetch(`https://registry.npmjs.org/${dep}`)
```
dominykas, line 80: We should use `pacote` for this, so that it can support custom registries. Happy to address this as part of #6.

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+  pkg = JSON.parse(pkg)
+  return pkg.name
+}
+
+const getGitHubOrgRepo = module.exports.getGitHubOrgRepo =
+async function getGitHubOrgRepo (dep) {
+  const urlRegex = /github.com\/([^/])+\/[^/]+/g
+  const resp = await fetchRegistryInfo(dep)
+  let org, repo
+  if (resp.repository && resp.repository.url) {
```
dominykas, line 63: `resp.repository` can also be a string (the URL). And in this case `git-url-parse` is definitely handy.

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+}
+
+const checkPackageInPackageJSON = module.exports.checkPackageInPackageJSON =
+function checkPackageInPackageJSON (dep, packageJSON) {
+  return Object.prototype.hasOwnProperty.call(packageJSON.dependencies, dep)
+}
+
+const getOrgRepo = module.exports.getOrgRepo =
+function getOrgRepo (url) {
+  const repoOrgArr = (url.split('github.com'))[1].split('/')
```
dominykas, line 33: I find https://www.npmjs.com/package/git-url-parse handy for things like this, esp. if we're going to support Github Enterprise.

andrewhughes101:
Thanks for the reviews @dominykas @BethGriggs, Im going to land this, any outstanding items in the reviews I'll turn into new issues so anyone can work in parallel on them.

bin/wiby.js
```diff
@@ -1,12 +1,26 @@
 #!/usr/bin/env node
 
+require('dotenv').config()
 const test = require('../lib/test')
 const result = require('../lib/result')
 
-if (process.argv[2] === 'test') {
-  test()
+const args = require('yargs')
+  .option('test', {
```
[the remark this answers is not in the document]
andrewhughes101: Covered by https://github.com/pkgjs/wiby/issues/7

lib/test.js
andrewhughes101: Covered by #9

lib/test.js
andrewhughes101: Covered by #10

lib/test.js
andrewhughes101: Covered by #10

PR #4, opened 2020-06-18
lib/result: add initial prototype for wiby --result

andrewhughes101:
Basic prototype of `wiby result` which will fetch the most recently completed ci runs for a dependent.

Usage: 
Run `wiby result --dependent dependent_url` from the parent packages directory (like wiby test) and the results for the latest ci runs on the wiby branch will be returned. This prototype assumes `wiby test` has already been run.

Sample output:
```sh
Tests on branch "wiby-i-should-pass"
{
  'build (14.x)': 'success',
  'build (12.x)': 'failure',
  'build (10.x)': 'cancelled'
}
```

### DIFF
--- a/bin/wiby.js
+++ b/bin/wiby.js
@@ -33,7 +33,10 @@ yargs
         })
     },
     (argv) => {
-      result(argv.dependent)
+      result(argv.dependent).catch(e => {
+        console.error(e)
+        process.exitCode = e.code || 1
+      })
     }
   )
   .help()
--- a/bin/wiby.js
+++ b/bin/wiby.js
@@ -31,6 +31,6 @@ yargs
   )
   .help()
   .strict()
-  .argv
+  .parse()
 
 // Usage: wiby test --dependent=URL
--- a/bin/wiby.js
+++ b/bin/wiby.js
@@ -24,9 +24,16 @@ yargs
   .command(
     'result',
     'Fetch the results of your tests',
-    () => {},
-    () => {
-      result()
+    (yargs) => {
+      yargs
+        .option('dependent', {
+          desc: 'URL of a dependent',
+          demandOption: true,
+          type: 'string'
+        })
+    },
+    (argv) => {
+      result(argv.dependent)
     }
   )
   .help()
@@ -34,3 +41,4 @@ yargs
   .parse()
 
 // Usage: wiby test --dependent=URL
+// Usage: wiby result --dependent=URL
--- a/lib/github.js
+++ b/lib/github.js
@@ -121,3 +121,19 @@ async function createBranch (owner, repo, commitSha, branch) {
     ref: `refs/heads/${branch}`
   })
 }
+
+module.exports.getChecks = async function getChecks (owner, repo, branch) {
+  return octokit.checks.listForRef({
+    owner,
+    repo,
+    ref: branch
+  })
+}
+
+module.exports.getCommitStatusesForRef = async function getCommitStatusesForRef (owner, repo, branch) {
+  return octokit.repos.listCommitStatusesForRef({
+    owner,
+    repo,
+    ref: branch
+  })
+}
--- a/lib/github.js
+++ b/lib/github.js
@@ -10,8 +10,7 @@ const graphqlWithAuth = graphql.defaults({
   }
 })
 
-module.exports.getPackageJson =
-async function getPackageJson (owner, repo) {
+module.exports.getPackageJson = async function getPackageJson (owner, repo) {
   try {
     const resp = await graphqlWithAuth({
       query: `query($owner: String!, $repo: String!) {
@@ -36,8 +35,7 @@ async function getPackageJson (owner, repo) {
   }
 }
 
-module.exports.getPermissions =
-async function getPermissions (owner, repo) {
+module.exports.getPermissions = async function getPermissions (owner, repo) {
   try {
     const resp = await graphqlWithAuth({
       query: `query($owner: String!, $repo: String!) {
@@ -58,8 +56,7 @@ async function getPermissions (owner, repo) {
   }
 }
 
-module.exports.getShas =
-async function getShas (owner, repo) {
+module.exports.getShas = async function getShas (owner, repo) {
   const resp = await octokit.repos.listCommits({
     owner,
     repo,
@@ -70,11 +67,8 @@ async function getShas (owner, repo) {
   return [headSha, treeSha]
 }
 
-module.exports.createBlob =
-async function createBlob (owner, repo, file) {
-  const {
-    data: { sha: blobSha }
-  } = await octokit.git.createBlob({
+module.exports.createBlob = async function createBlob (owner, repo, file) {
+  const { data: { sha: blobSha } } = await octokit.git.createBlob({
     owner,
     repo,
     content: file,
@@ -84,7 +78,7 @@ async function createBlob (owner, repo, file) {
 }
 
 module.exports.createTree = async function createTree (owner, repo, treeSha, blobSha) {
-  const resp = await octokit.git.createTree({
+  const { data: { sha: newTreeSha } } = await octokit.git.createTree({
     owner,
     repo,
     base_tree: treeSha,
@@ -95,25 +89,21 @@ module.exports.createTree = async function createTree (owner, repo, treeSha, blo
       Accept: 'application/json'
     }
   })
-  const newTreeSha = resp.data.sha
   return newTreeSha
 }
 
-module.exports.createCommit =
-async function createCommit (owner, repo, message, newTreeSha, headSha) {
-  const resp = await octokit.git.createCommit({
+module.exports.createCommit = async function createCommit (owner, repo, message, newTreeSha, headSha) {
+  const { data: { sha: commitSha } } = await octokit.git.createCommit({
     owner,
     repo,
     message: message,
     tree: newTreeSha,
     parents: [headSha]
   })
-  const commitSha = resp.data.sha
   return commitSha
 }
 
-module.exports.createBranch =
-async function createBranch (owner, repo, commitSha, branch) {
+module.exports.createBranch = async function createBranch (owner, repo, commitSha, branch) {
   await octokit.git.createRef({
     owner,
     repo,
--- a/lib/result.js
+++ b/lib/result.js
@@ -27,8 +27,7 @@ module.exports = async function (url) {
 }
 
 const getBranchName = module.exports.getBranchName = async function getBranchName (dep) {
-  const branch = `wiby-${dep}`
-  return branch
+  return `wiby-${dep}`
 }
 
 const getResultForEachRun = module.exports.getResultForEachRun = function getResultForEachRun (runs) {
--- a/lib/result.js
+++ b/lib/result.js
@@ -1,3 +1,44 @@
-module.exports = () => {
-  console.log('wiby result')
+const test = require('./test')
+const github = require('./github')
+const gitURLParse = require('git-url-parse')
+
+module.exports = async function (url) {
+  const parentPkgJSON = await test.getLocalPackageJSON()
+  const parentPkgInfo = gitURLParse(parentPkgJSON.repository.url)
+  console.log(`Parent module: ${parentPkgInfo.owner}/${parentPkgJSON.name}`)
+
+  const dependentPkgInfo = gitURLParse(url)
+  const dependentPkgJSON = await github.getPackageJson(dependentPkgInfo.owner, dependentPkgInfo.name)
+  console.log(`Dependent module: ${dependentPkgInfo.owner}/${dependentPkgInfo.name}`)
+
+  if (!test.checkPackageInPackageJSON(parentPkgJSON.name, dependentPkgJSON)) {
+    throw new Error('Dependency not found in package.json')
+  }
+
+  const branch = await getBranchName(parentPkgJSON.name)
+  let resp = await github.getChecks(dependentPkgInfo.owner, dependentPkgInfo.name, branch)
+  if (resp.data.check_runs.length === 0) {
+    resp = await github.getCommitStatusesForRef(dependentPkgInfo.owner, dependentPkgInfo.name, branch)
+  }
+  const runs = resp.data.check_runs || resp.data
+  console.log(`Tests on branch "${branch}"`)
+  const results = getResultForEachRun(runs)
+  console.log(results)
+}
+
+const getBranchName = module.exports.getBranchName = async function getBranchName (dep) {
+  const branch = `wiby-${dep}`
+  return branch
+}
+
+const getResultForEachRun = module.exports.getResultForEachRun = function getResultForEachRun (runs) {
+  const results = []
+  runs.forEach((check) => {
+    if (check.status === 'queued') {
+      results.push([check.name, check.status])
+    } else {
+      results.push([check.name, check.conclusion])
+    }
+  })
+  return results
 }
--- a/lib/result.js
+++ b/lib/result.js
@@ -12,7 +12,7 @@ module.exports = async function (url) {
   console.log(`Dependent module: ${dependentPkgInfo.owner}/${dependentPkgInfo.name}`)
 
   if (!test.checkPackageInPackageJSON(parentPkgJSON.name, dependentPkgJSON)) {
-    throw new Error('Dependency not found in package.json')
+    throw new Error(`${parentPkgInfo.owner}/${parentPkgJSON.name} not found in the package.json of ${dependentPkgInfo.owner}/${dependentPkgInfo.name}`)
   }
 
   const branch = await getBranchName(parentPkgJSON.name)
--- a/lib/result.js
+++ b/lib/result.js
@@ -31,13 +31,10 @@ const getBranchName = module.exports.getBranchName = async function getBranchNam
 }
 
 const getResultForEachRun = module.exports.getResultForEachRun = function getResultForEachRun (runs) {
-  const results = []
-  runs.forEach((check) => {
+  return runs.map((check) => {
     if (check.status === 'queued') {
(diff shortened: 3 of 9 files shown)


lib/github.js
```diff
@@ -122,3 +121,13 @@ async function createBranch (owner, repo, commitSha, branch) {
...
   })
 }
+
+module.exports.getChecks =
+async function getChecks (owner, repo, branch) {
+  const resp = await octokit.checks.listForRef({
+    owner,
+    repo,
+    ref: branch
+  }).catch(e => { throw (e) })
```
dominykas, line 131: Is this `catch()` necessary? If you just rethrow the error - it's the same as not having the `catch()` at all?

lib/result.js
```diff
@@ -1,3 +1,32 @@
...
+  const [org, repo] = await test.getGitHubOrgRepo(pkg).catch(err => { throw (err) })
+  const resp = await github.getChecks(org, repo, branch).catch(err => { throw (err) })
+  const runs = resp.data.check_runs
+  console.log(`Tests on branch "${branch}"`)
+  const resultJSON = await getResultForEachRun(runs).catch(err => { throw (err) })
+  console.log(resultJSON)
+}
+
+const getBranchName =
+module.exports.getBranchName = async function getBranchName () {
```
dominykas, line 15: This is reusable inside `test.pushPatch()`, right?

lib/result.js
```diff
@@ -1,3 +1,32 @@
...
+  const dep = await test.getLocalPackageName().catch(err => { throw (err) })
+  const branch = 'wiby-' + dep
+  return branch
+}
+
+const getResultForEachRun =
+module.exports.getResultForEachRun = async function getResultForEachRun (runs) {
+  const results = {}
```
dominykas, line 23: Is there a specific reason to use an object here? I figure keeping this as an array may be a bit more versatile (mostly cause I have no idea if GH enforces the `name` to be unique)

lib/github.js
```diff
@@ -122,3 +121,13 @@ async function createBranch (owner, repo, commitSha, branch) {
     ref: `refs/heads/${branch}`
   })
 }
+
+module.exports.getChecks =
+async function getChecks (owner, repo, branch) {
+  const resp = await octokit.checks.listForRef({
```
dominykas, line 127: Should this also read from `octokit.repos.listCommitStatusesForRef`? Although I'll admit I'm not too familiar with these APIs. I can see the checks empty on some of my older repos, and the statuses filled in, but I'm having trouble retrieving either on our GH Enterperise instance. I wonder if there's documentation/some comments on this somewhere 🤔

bin/wiby.js
```diff
@@ -1,12 +1,28 @@
...
+  })
+  .option('result', {
+    alias: 'r',
+    describe: 'Get the result of your tests'
+  })
+  .option('package', {
+    alias: 'pkg',
+    describe: 'Specific dependent to find test results for'
+  })
+  .argv
```
ljharb, line 20:
```suggestion
  .strict()
  .help()
  .parse()
```
the "strict" means it will show the help when an invalid command is supplied; help enables the help; and .parse() both would throw if you made a typo instead of silently being undefined.

bin/wiby.js
```diff
@@ -1,12 +1,28 @@
...
+
+if (args.test) {
+  test(args.test)
 }
 
-if (process.argv[2] === 'result') {
-  result()
+if (args.result && args.package) {
+  result(args.package).catch(e => { console.log(e) })
```
ljharb, line 27:
```suggestion
  result(args.package).catch(e => {
    console.error(e);
    process.exitCode = e.code || 1;
  })
```

lib/github.js
```diff
@@ -0,0 +1,133 @@
...
+})
+
+const graphqlWithAuth = graphql.defaults({
+  headers: {
+    authorization: `token ${process.env.GITHUB_TOKEN}`
+  }
+})
+
+module.exports.getPackageJson =
+async function getPackageJson (owner, repo) {
```
ljharb, line 14:
```suggestion
module.exports.getPackageJson = async function getPackageJson (owner, repo) {
```
minor nit: it's awkward to have a line break after an `=`. same throughout.

lib/github.js
```diff
@@ -0,0 +1,133 @@
...
+async function createCommit (owner, repo, message, newTreeSha, headSha) {
+  const resp = await octokit.git.createCommit({
+    owner,
+    repo,
+    message: message,
+    tree: newTreeSha,
+    parents: [headSha]
+  })
+  const commitSha = resp.data.sha
+  return commitSha
```
ljharb, line 112:
```suggestion
  return { data: { sha } }  = await octokit.git.createCommit({
    owner,
    repo,
    message: message,
    tree: newTreeSha,
    parents: [headSha]
  })
```

lib/github.js
```diff
@@ -0,0 +1,133 @@
...
+}
+
+module.exports.getChecks =
+async function getChecks (owner, repo, branch) {
+  const resp = await octokit.checks.listForRef({
+    owner,
+    repo,
+    ref: branch
+  }).catch(e => { throw (e) })
+  return resp
```
ljharb, line 132:
```suggestion
  return octokit.checks.listForRef({
    owner,
    repo,
    ref: branch
  })
```

lib/result.js
```diff
@@ -1,3 +1,32 @@
-module.exports = () => {
-  console.log('wiby result')
+const test = require('./test')
+const github = require('./github')
+
+module.exports = async function (pkg) {
+  const branch = await getBranchName()
+  const [org, repo] = await test.getGitHubOrgRepo(pkg).catch(err => { throw (err) })
+  const resp = await github.getChecks(org, repo, branch).catch(err => { throw (err) })
```
ljharb, line 7:
```suggestion
  const [org, repo] = await test.getGitHubOrgRepo(pkg)
  const resp = await github.getChecks(org, repo, branch)
```
a catch that just re-throws is redundant, and will also break node's ability to show the line number from where it originated.

lib/result.js
```diff
@@ -1,3 +1,32 @@
...
+  const runs = resp.data.check_runs
+  console.log(`Tests on branch "${branch}"`)
+  const resultJSON = await getResultForEachRun(runs).catch(err => { throw (err) })
```
ljharb, line 10:
```suggestion
  const resultJSON = await getResultForEachRun(runs)
```

lib/result.js
```diff
@@ -1,3 +1,32 @@
...
+  console.log(resultJSON)
+}
+
+const getBranchName =
+module.exports.getBranchName = async function getBranchName () {
+  const dep = await test.getLocalPackageName().catch(err => { throw (err) })
```
ljharb, line 16:
```suggestion
  const dep = await test.getLocalPackageName()
```

lib/result.js
```diff
@@ -1,3 +1,32 @@
...
+  const branch = 'wiby-' + dep
```
ljharb, line 17:
```suggestion
  const branch = `wiby-${dep}`
```

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+  }
+  const patch = await getPatch(dep)
+  console.log('Created patch: ', patch)
+  const patchedPackageJSON = applyPatch(patch, dep, packageJSON)
+  await pushPatch(patchedPackageJSON, org, repo, dep)
+}
+
+const getCommitHash = module.exports.getCommitHash =
+async function getCommitHash (owner, repo) {
+  const headSha = (await github.getShas(owner, repo))[0]
```
ljharb, line 22:
```suggestion
  const [headSha] = await github.getShas(owner, repo)
```

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+const checkPackageInPackageJSON = module.exports.checkPackageInPackageJSON =
+function checkPackageInPackageJSON (dep, packageJSON) {
+  return Object.prototype.hasOwnProperty.call(packageJSON.dependencies, dep)
+}
+
+const getOrgRepo = module.exports.getOrgRepo =
+function getOrgRepo (url) {
+  const repoOrgArr = (url.split('github.com'))[1].split('/')
+  const org = repoOrgArr[1]
+  const repo = repoOrgArr[2]
```
ljharb, line 35:
```suggestion
  const [, org, repo] = url.split('github.com')[1].split('/')
```

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+}
+
+const getLocalPackageName = module.exports.getLocalPackageName =
+async function getLocalPackageName (pkgPath) {
+  pkgPath = pkgPath || 'package.json'
+  let pkg = await fsPromises.readFile(pkgPath).catch((err) => {
+    throw (err)
+  })
+  pkg = JSON.parse(pkg)
+  return pkg.name
```
ljharb, line 55:
```suggestion
  const pkg = JSON.parse(await fsPromises.readFile(pkgPath))
  return pkg.name
```

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+}
+
+const getGitHubOrgRepo = module.exports.getGitHubOrgRepo =
+async function getGitHubOrgRepo (dep) {
+  const urlRegex = /github.com\/([^/])+\/[^/]+/g
+  const resp = await fetchRegistryInfo(dep)
+  let org, repo
+  if (resp.repository && resp.repository.url) {
+    let gitUrl = (resp.repository.url).match(urlRegex)
```
ljharb, line 64:
```suggestion
    let gitUrl = resp.repository.url.match(urlRegex)
```

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+    if (gitUrl) {
+      gitUrl = gitUrl[0].replace(/(\.git)/g, '')
+      org = getOrgRepo(gitUrl)[0]
+      repo = getOrgRepo(gitUrl)[1]
```
ljharb, line 68:
```suggestion
      [org, repo] = getOrgRepo(gitUrl)
```

lib/test.js
```diff
@@ -1,3 +1,105 @@
...
+    }
+  }
+  if (!org && !repo) {
+    org = 'undefined'
+    repo = 'undefined'
```
ljharb, line 73: this seems very weird, especially since https://github.com/undefined is a thing?

lib/github.js
dominykas: I specifically did not pick on this and other nits, as this should all be covered by the linters? I assume half of that is `standard`... But I don't use that style, so maybe I'm wrong. Regardless - we should just add an `.eslintrc`, discuss the actual style in a separate issue if we have to and be done with it, rather than pick on this in prototype PRs.

ljharb: That's generally fair, but I'm not sure there's a linter rule that properly blocks line breaks after `=`, which is why i commented on this one (and not on all the missing semicolons :-p)

lib/github.js
andrewhughes101: Fixed

lib/result.js
andrewhughes101: Not sure if GH does enforce the name to be unique. Switched to using an array for the results.

lib/github.js
andrewhughes101: I cant get octokit.repos.listCommitStatusesForRef to return anything when supplying a branch name. So I cant compare the endpoints at the moment. I'll try again tomorrow but for now have not updated PR.

bin/wiby.js
andrewhughes101: Fixed

bin/wiby.js
andrewhughes101: Fixed

lib/github.js
andrewhughes101: Fixed throughout

lib/github.js
andrewhughes101: Fixed

lib/result.js
andrewhughes101: Fixed

lib/result.js
andrewhughes101: Fixed

lib/result.js
andrewhughes101: Fixed

lib/result.js
andrewhughes101: Fixed

lib/test.js
andrewhughes101: Fixed

lib/test.js
andrewhughes101: Fixed

lib/test.js
andrewhughes101: Fixed

lib/test.js
andrewhughes101: Fixed

lib/test.js
andrewhughes101: Fixed

lib/test.js
andrewhughes101: I think I was going to add some error handling for repos/orgs that didnt exist, but never added it in so think it makes sense to just remove this, especially as https://github.com/undefined is a thing.

lib/github.js
andrewhughes101: This is returning the entire response object instead of just the sha, unless I'm missing something. Not sure if we need to return the entire object.

lib/result.js
andrewhughes101: Yes I think it is, I'll add it in thanks!

lib/test.js
ljharb: it would work fine as a literal `undefined`, i think, just not a string

lib/github.js
ljharb: whoops, you're right, let me fix the suggestion.

andrewhughes101: Fixed

lib/result.js
andrewhughes101: Fixed

lib/github.js
dominykas: I think it depends on the repo and on the way the integrations were set up. Here's an old repo of mine, which does return stuff for the status API: https://api.github.com/repos/dominykas/grunt-checkbranch/commits/5352d46f3c0c370863cf6889afb410a60ca078e8/status

It does not return anything for the checks though...

andrewhughes101: Think I've managed to fix it. Was seeing some weirdness where @octokit/rest didnt think octokit.repos.listCommitStatusesForRef was valid. 

I've included it now, it will try to use ocotkit.checks.listForRef first but fallback to octokit.repos.listCommitStatusesForRef if no checks were returned.

dominykas:
Code looks good, although I haven't tested it just yet - this now has conflicts (sorry).

andrewhughes101:
Fixed conflicts. The result flow has changed slightly so that you now pass the url of a dependent instead of just the dependents package name. This means the result flow will work with packages that are not published to npm.

lib/result.js
```diff
@@ -1,3 +1,44 @@
...
+  const parentPkgJSON = await test.getLocalPackageJSON()
+  const parentPkgInfo = gitURLParse(parentPkgJSON.repository.url)
+  console.log(`Parent module: ${parentPkgInfo.owner}/${parentPkgJSON.name}`)
+
+  const dependentPkgInfo = gitURLParse(url)
+  const dependentPkgJSON = await github.getPackageJson(dependentPkgInfo.owner, dependentPkgInfo.name)
+  console.log(`Dependent module: ${dependentPkgInfo.owner}/${dependentPkgInfo.name}`)
+
+  if (!test.checkPackageInPackageJSON(parentPkgJSON.name, dependentPkgJSON)) {
+    throw new Error('Dependency not found in package.json')
```
dominykas, line 15:
```suggestion
    throw new Error(`${parentPkgInfo.owner}/${parentPkgJSON.name} not found in the package.json of ${dependentPkgInfo.owner}/${dependentPkgInfo.name}`)
```

andrewhughes101:
@ljharb So the main reason I changed it to use a URL was because the current `wiby test` flow is using a URL and I know this makes it easier to try this tool out on a parent and dependent github repo that havent been published to npm.

I'm happy to change this to accept "whatever npm accepts" but I think that should also be applied to `wiby test` so I'm wondering would it be best to change both in a different PR, or change it here and open another PR changing `wiby test`

dominykas:
>  would it be best to change both in a different PR

I very much agree that we should accept more formats, but I also very much agree that this needs to go into a separate PR, primarily because this PR completes the MVP-est wiby - we do need both components `test` and `result` before we can build further.

lib/result.js
```diff
@@ -1,3 +1,44 @@
...
+  }
+  const runs = resp.data.check_runs || resp.data
+  console.log(`Tests on branch "${branch}"`)
+  const results = getResultForEachRun(runs)
+  console.log(results)
+}
+
+const getBranchName = module.exports.getBranchName = async function getBranchName (dep) {
+  const branch = `wiby-${dep}`
+  return branch
```
ljharb, line 31:
```suggestion
  return `wiby-${dep}`
```
or does this need a TODO comment instead?

lib/result.js
```diff
@@ -1,3 +1,44 @@
...
+const getResultForEachRun = module.exports.getResultForEachRun = function getResultForEachRun (runs) {
+  const results = []
+  runs.forEach((check) => {
+    if (check.status === 'queued') {
+      results.push([check.name, check.status])
+    } else {
+      results.push([check.name, check.conclusion])
+    }
+  })
+  return results
```
ljharb, line 43:
```suggestion
  return runs.map((check) => {
    if (check.status === 'queued') {
      return [check.name, check.status]
    }
    return [check.name, check.conclusion]
  })
```

lib/result.js
andrewhughes101: I think this is fine for now, I think we should open an issue around branches/refs and using version numbers in the branches as well

dominykas:
* [x] No blocking reviews

PR #12, opened 2020-07-05
Add CI

### LINKED ISSUE #5: Add CI
I'm more familiar with Travis - happy to PR that.

dominykas:
Some CI is better than no CI

Closes #5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,9 @@
 language: node_js
-sudo: false
 
 node_js:
-  - "10"
-  - "12"
   - "14"
+  - "12"
+  - "10"
 
 cache:
   npm: false
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: node_js
+sudo: false
+
+node_js:
+  - "10"
+  - "12"
+  - "14"
+
+cache:
+  npm: false
+
--- a/test/github.js
+++ b/test/github.js
@@ -3,11 +3,13 @@ const tap = require('tap')
 const github = require('../lib/github')
 const CONFIG = require('./fixtures/config')
 
-tap.test('package.json can be fetched with a valid url', async tap => {
-  tap.equal(JSON.stringify(await github.getPackageJson(CONFIG.DEP_ORG, CONFIG.DEP_REPO)), JSON.stringify(CONFIG.PKGJSON))
-  // tap.throws( await pkgTest.getPackageJson('not-an-org', 'not-a-repo'))
-})
+if (process.env.GITHUB_TOKEN) {
+  tap.test('package.json can be fetched with a valid url', async tap => {
+    tap.equal(JSON.stringify(await github.getPackageJson(CONFIG.DEP_ORG, CONFIG.DEP_REPO)), JSON.stringify(CONFIG.PKGJSON))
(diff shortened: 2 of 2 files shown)


dominykas:
Will add my token for tests to work in CI, but @andrewhughes101 will need to add me as admin on the test repo.

Ideally, the tests would pass without any setup (i.e. without the need for a token), so I guess we should split the tests that need network and the ones that don't at some point later.

.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: node_js
+sudo: false
+
+node_js:
+  - "10"
+  - "12"
+  - "14"
+
+cache:
+  npm: false
```
ljharb, line 10: this is the default, fwiw, so this line isn't needed

.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: node_js
+sudo: false
+
+node_js:
+  - "10"
+  - "12"
+  - "14"
```
ljharb, line 7: nit: sort the newest at the top
```suggestion
  - "14"
  - "12"
  - "10"
```

test/github.js
```diff
@@ -3,11 +3,13 @@ const tap = require('tap')
 const github = require('../lib/github')
 const CONFIG = require('./fixtures/config')
 
-tap.test('package.json can be fetched with a valid url', async tap => {
-  tap.equal(JSON.stringify(await github.getPackageJson(CONFIG.DEP_ORG, CONFIG.DEP_REPO)), JSON.stringify(CONFIG.PKGJSON))
-  // tap.throws( await pkgTest.getPackageJson('not-an-org', 'not-a-repo'))
-})
+if (process.env.GITHUB_TOKEN) {
```
ljharb, line 6: does tap not have a "skip" mechanism?

.travis.yml
dominykas: So the docs are wrong? https://docs.travis-ci.com/user/caching/#npm-cache

test/github.js
dominykas: My google led me to issues for `jest`, not `tap` 🤦 https://github.com/pkgjs/wiby/pull/12/commits/0d5460dcb124a38f1ab48f9c1e13eac07235c194

.travis.yml
dominykas: Updated in https://github.com/pkgjs/wiby/pull/12/commits/5b5e684f1704ea1bf553f717f69211a902e1f21e, but for the record I find this change entirely pointless.

Nits like this are distracting, discouraging and demotivating.

.travis.yml
ljharb: apparently i'm wrong as of 6 days ago :-) "Please note that as of July 2019, npm is cached by default on Travis CI" my mistake.

.travis.yml
ljharb: I'm sorry I caused you to feel this way; I'd love to understand more about why it's discouraging or demotivating. To me, this is just part of any code review. It being a "nit" just means it's not a blocker, and i'd be happy to explain why it's not pointless if you're interested.

dominykas: The motivation for this probably matters more on https://github.com/nodejs/ci-config-travis/pull/6 than here.

ljharb: In an imported config, the order is backwards; so i believe it's correct there :-)

PR #14, opened 2020-07-20
lib: handle not published packages

BethGriggs:
This PR removes the need for modules to be published to npm before they can be tested. This involved a bit more refactoring than I first anticipated...
 
To summarise, this PR: 
- Introduces `git-url-parse` module
- Extracts the parent module repository from the local `package.json` using the `git-url-parse` module.
  - This enables us to remove `getGitHubOrgRepo`, `getOrgRepo`, and `fetchRegistryInfo` functions.
- Renames variables such as `dep`
  - These changes were initially to just to help me understand the flow of the program, but I think it is worth keeping more specific naming - although happy to change these. 
- Changes the `getPatch` function name to `getCommitURL`
  - Open to bike-shedding on this function name. I felt it should be changed as when I was working through the code I assumed that the function returned a Git Patch, but it returns the URL that we wish to patch the `package.json` with.
- Removes the tests for the removed functions. Updates the test for `getPatch` where the function has been renamed.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -1,20 +1,25 @@
 const fsPromises = require('fs').promises
-const fetch = require('node-fetch')
 const github = require('./github')
+const gitURLParse = require('git-url-parse')
 
 module.exports = async function (url) {
-  const dep = await getLocalPackageName()
-  console.log('Got local package name: ', dep)
-  const [org, repo] = getOrgRepo(url)
-  console.log(`Got org: ${org} and repo: ${repo}`)
-  const packageJSON = await github.getPackageJson(org, repo)
-  if (!checkPackageInPackageJSON(dep, packageJSON)) {
+  const parentPkgJSON = await getLocalPackageJSON()
+  const parentPkgInfo = gitURLParse(parentPkgJSON.repository.url)
+  console.log(`Parent module: ${parentPkgInfo.owner}/${parentPkgJSON.name}`)
+
+  const dependentPkgInfo = gitURLParse(url)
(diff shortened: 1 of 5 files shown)


BethGriggs:
It might be easier to review the file in full - https://github.com/pkgjs/wiby/blob/5a124183b53d432170315b541d864a3cdae12a88/lib/test.js

BethGriggs:
Thanks @rodion-arr,  I've addressed the unused var.

I think there's an ongoing discussion about our choice of linter, so I may defer adding standard to the test script for a later PR.

lib/test.js
```diff
@@ -1,20 +1,25 @@
...
+  const parentPkgJSON = await getLocalPackageJSON()
+  const parentPkgInfo = gitURLParse(parentPkgJSON.repository.url)
+  console.log(`Parent module: ${parentPkgInfo.owner}/${parentPkgJSON.name}`)
+
+  const dependentPkgInfo = gitURLParse(url)
+  const dependentPkgJSON = await github.getPackageJson(dependentPkgInfo.owner, dependentPkgInfo.name)
+  console.log(`Dependent module: ${dependentPkgInfo.owner}/${dependentPkgInfo.name}`)
+
+  if (!checkPackageInPackageJSON(parentPkgJSON.name, dependentPkgJSON)) {
     throw new Error('Dependency not found in package.json')
```
dominykas, line 15: Not a part of this PR, but still:

```suggestion
    throw new Error(`${parentPkgInfo.owner}/${parentPkgJSON.name} not found in package.json of ${dependentPkgInfo.owner}/${dependentPkgInfo.name}`)
```

dominykas:
* [x] No blocking reviews

Issue #7, opened 2020-06-29
Use a Yargs command for `wiby test` 


BethGriggs:
As per https://github.com/pkgjs/wiby/pull/3#discussion_r446881882

PR #17, opened 2020-07-30
bin: replace .option with .command

AshCripps:
fixes: https://github.com/pkgjs/wiby/issues/7


dominykas:
* [x] No blocking reviews
* [ ] 72 hours since the PR was openened

PR #22, opened 2020-08-01
Update unit tests to use wiby-test organization

### LINKED ISSUE #21: Use a shared organization for tests
I know someone mentioned this earlier at some point, but either way - I went ahead and created https://github.com/wiby-test and invited the same people who are contributors/owners here. I recommend you unsubscribe from notifications there :)

I figure a separate org is better for this, to avoid polluting pkgjs with repositories which are only used for automated tests. Don't think we need any kind of policies around these repositories - the intent is to just close everything that happens there?

I plan to next scaffold out some test repos there, which will always produce the same test results, so that we can use them to test wiby end-to-end (there's already some repos being accessed in the unit tests, but they're under @andrewhughes101's account, so I'll replace those).

dominykas:
Closes #21


dominykas:
* [x] No blocking reviews

PR #25, opened 2020-08-03
Update CI

### LINKED ISSUE #18: Enable linting as part of `npm test`
First mentioned in https://github.com/pkgjs/wiby/pull/14#pullrequestreview-451685315:

> And it will be great to add `standard` command run to `test` script

While I wouldn't call `standard` "great" per se :trollface:, we might as well add it to the `npm test`, unless there's any objections or we plan to change the style?

rodion-arr:
- Add separate `lint` and `test-only` npm scripts
- Update CI to run `lint` once and then run tests on matrix

Closes #18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
 jobs:
   fast_finish: true
   include:
-    - node_js: 14
+    - node_js: lts/*
       env: LINT=true
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,11 @@ node_js:
 
 env:
   - TESTS=true
-  - LINT=true
 
 jobs:
   fast_finish: true
-  exclude:
-    - node_js: 10
-      env: LINT=true
-    - node_js: 12
+  include:
+    - node_js: 14
       env: LINT=true
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,21 @@ node_js:
   - "12"
   - "10"
 
+env:
+  - TESTS=true
+  - LINT=true
+
+jobs:
+  fast_finish: true
+  exclude:
+    - node_js: 10
+      env: LINT=true
+    - node_js: 12
+      env: LINT=true
+
+script:
+  - 'if [ -n "${LINT-}" ]; then npm run pretest ; fi'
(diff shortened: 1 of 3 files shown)


rodion-arr:
Hm.. strange, `standard` is passing locally but failed on CI.. will take a look

rodion-arr:
I wasn't rebased on latest upstream. Linting on `master` is failing

bin/wiby.js
```diff
@@ -31,6 +31,5 @@ yargs
   )
   .help()
   .strict()
```
dominykas, line 33:
```suggestion
  .strict()
  .parse()
```

The `.argv` getter performs parsing - it is the thing that executes the actual commands - without either `.argv` or `.parse()` things don't work.

.travis.yml
```diff
@@ -5,5 +5,21 @@ node_js:
...
+  - TESTS=true
+  - LINT=true
+
+jobs:
+  fast_finish: true
+  exclude:
+    - node_js: 10
+      env: LINT=true
+    - node_js: 12
+      env: LINT=true
```
dominykas, line 18: The `exclude` list will require maintainenance each time a new version is added (but also I'd like to replace the `node_js` bit with https://github.com/nodejs/ci-config-travis as soon as it gets merged to avoid having to maintain the version list here separately) - can we instead add an explicit, separate job just for linting via `include`?

ljharb: yes, that's the only way the linting job should ever be included :-)

bin/wiby.js
rodion-arr: thanks!

.travis.yml
rodion-arr: Thanks for suggestion, this will be much prettier 👍

.travis.yml
```diff
@@ -5,5 +5,18 @@ node_js:
   - "12"
   - "10"
 
+env:
+  - TESTS=true
+
+jobs:
+  fast_finish: true
+  include:
+    - node_js: 14
```
ljharb, line 14:
```suggestion
    - node_js: lts/*
```
no need to tie it to 14; this way it'll just always run on the latest LTS.

PR #27, opened 2020-08-04
doc: Update README.md for token and commands

AshCripps:
Update the README to properly explain how the tool works.

fixes: https://github.com/pkgjs/wiby/issues/16


README.md
```diff
@@ -4,3 +4,30 @@
...
+
+usage: `wiby test --dependent=URL_TO_DEPENDENT`
+
+Use this command to test your breaking changes against any one of your dependent. Wiby will go off to the dependents repo and create a branch with a patch to the `package.json` pointing to your latest version (with the new changes) triggering the dependents CI to run.
+
+### result
+
+usage: `wiby result --dependent=URL_TO_DEPENDENT`
+
+Use this command to fetch the results of your latest test against a dependent.
```
AshCripps, line 33: Would we want a sample output here aswell? ive left result a bit more sparse as the functionality hasnt landed yet.

README.md
```diff
@@ -4,3 +4,30 @@
...
+
+Wiby requires an environment variable `GITHUB_TOKEN` set to a github access token.
+
+Example: `export GITHUB_TOKEN=XXXXX`
+
+For more information on creating a github token see [Github's docs](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token)
+
+### Running location
+
+Wiby is designed to be run from inside the folder containing your source code.
```
AshCripps, line 19: Am I 100% right here?

README.md
```diff
@@ -4,3 +4,30 @@
 
 This repository is managed by the [Package Maintenance Working Group](https://github.com/nodejs/package-maintenance), see [Governance](https://github.com/nodejs/package-maintenance/blob/master/Governance.md).
 
+## Pre-requisites
+
+### Github Token
+
+Wiby requires an environment variable `GITHUB_TOKEN` set to a github access token.
```
dominykas, line 11: I would a note that the token needs to grant push permissions to the dependent repos (at least until we implement something more sophisticated)

README.md
```diff
@@ -4,3 +4,30 @@
...
+
+Example: `export GITHUB_TOKEN=XXXXX`
+
+For more information on creating a github token see [Github's docs](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token)
+
+### Running location
+
+Wiby is designed to be run from inside the folder containing your source code.
+
+## Available commands
```
dominykas, line 21: I was thinking we could just have a list of links to `USAGE.md` in this section, and see if we can generate the `USAGE.md` out of yargs configuration? At the very simplest form it would start off with `wiby test --help >USAGE.md`.

The major benefit is that this would force us to include all the relevant information in the CLI docs, which is quite useful for the end users.

README.md
dominykas: At this point we just dump the API response, so possibly we should wait until we have something a little nicer?

README.md
dominykas: Kind of. The folder needs to be a git repository, and the `package.json` needs to contain the information about the source of the package on Github.

README.md
AshCripps: Ive added a `USAGE.md` file, but ive taken the explanation out - do we want it back or do we want to make the description more verbose on the command line to have consistent information? Also to create the `USAGE.md` I ran the commands manually with `--help` which is easy enough now but hopefully theres a easier way to create it as the project evolves with more options

dominykas: > do we want to make the description more verbose on the command line

Yes, I was thinking ☝️ this.

AshCripps: @dominykas in that case would it make sense to land this as is and then I'll open up another PR to make the CLI description more verbose and update `USAGE.md` or should I add that to this PR?

dominykas: I suppose it's fine - no reason to block it.

dominykas:
Needs an update with master for tests to pass.

Issue #26, opened 2020-08-03
Add tests for CLI flow


rodion-arr:
Current test suites does not cover CLI flow. As we see here (https://github.com/pkgjs/wiby/pull/25#discussion_r464467923) breaking `yargs` was not reflected in tests - all was green.

Will adding tests for CLI interface be valuable at current project phase?

dominykas:
I think it's useful, yes - we're also lacking some e2e coverage in general. Not sure how GH will react if we start running tests agains the live version of the repos from our local dev environments (and I was reluctant to add a GH token to the Travis jobs just yet - I figure we'll want to set up a bot account or smth for that - unsure about best practices there).

rodion-arr:
Why can't we just mock external requests in unit tests?

dominykas:
We can, but that requires a bit more work to test the CLI? Not that we shouldn't do it.

rodion-arr:
I'm going to write some more tests for CLI interface with mocked requests to GitHub. So at least it will cover expected commands and options but the real logic of this commands will be covered it theirs own tests. Is it OK from your side?

dominykas:
At the current stage, I'm more likely than not to approve any kind of tests ;)

PR #29, opened 2020-08-04
Fix tests

dominykas:
I didn't realize what the test in question was testing, so I just blindly updated the sha in #22. We need to come up with a better way to approach this.

The head of PKG_ORG / PKG_REPO will be changing often, so this test is bound to fail often. DEP_REPO, in theory, should have less frequent changes, so this test should be more stable until we come up with a better solution.

This probably qualifies for a speedy merge, as the master is red right now?


rodion-arr:
Hi @dominykas, can we land this? I'd to submit a PR for #26 after these fixes

PR #30, opened 2020-08-07
Add tests for CLI interface

### LINKED ISSUE #26: Add tests for CLI flow

rodion-arr:
- bin script refactored to be able to run cli and get output from tests
- calls to actual commands logic are mocked (only commands and options interfaces are under test)
- add `sinon` as mocking lib

Closes #26


lib/test.js
```diff
@@ -2,7 +2,7 @@ const fsPromises = require('fs').promises
 const github = require('./github')
 const gitURLParse = require('git-url-parse')
 
-module.exports = async function (url) {
+module.exports.test = async function (url) {
```
ljharb, line 5: why this change?

rodion-arr: I did it because Sinon couldn't replace this function with a mock

rodion-arr:
@ljharb The first idea was to invoke bin with 'exec' like it was implemented in 'support' but in this case I don't see possibilities to inject mocks

lib/test.js
ljharb: it seems very unfortunate to alter production code (and make requiring of this module much less ergonomic) just for testing.

ljharb:
What mocks are needed? Could we use something like `nock` to handle it?

rodion-arr:
General idea was to check that relevant function was called without really invoking it and do real requests to GitHub API while just checking that all passed to CLI options are correct.

https://github.com/rodion-arr/wiby/commit/091fc1fe5b75c201d066abc861a12c086d23d704#diff-583d616f4b01bf1fa153e71c0cef7de7R26-R29

ljharb:
it seems like we should be able to mock out the network requests with something like nock or vcr, and then exec the binary directly?

rodion-arr:
I got the point, I'll rollback the bin changes and try another approach. Thanks for suggestions 🤝

rodion-arr:
@ljharb, I tried `nock` and it is not working with child processes (https://github.com/nock/nock/issues/990)
Could you please send a link for what you called `vcr`, so I can try those one?
I'm going to play with this package - https://www.npmjs.com/package/node-vcr

ljharb:
ah, vcr is a rails thing, and i think nock is what i had in mind. however, you can pass nock env vars into NODE_OPTIONS, which the child process's `node` should respect, including `--require`?

rodion-arr:
Thanks, will try

lib/test.js
rodion-arr: reverted

rodion-arr:
@dominykas could you please rerun tests? It was failed due to rate limits

test/cli.js
```diff
@@ -0,0 +1,51 @@
...
+  })
+})
+
+tap.test('result command', async (tap) => {
+  tap.test('result command should require dependent option', async (tap) => {
+    try {
+      childProcess.execSync(`${wibyCommand} result`, { cwd: cwd }).toString()
+      tap.fail()
+    } catch (err) {
+      tap.equal(true, err.message.indexOf('Missing required argument: dependent') > 0)
```
ljharb, line 36:
```suggestion
      tap.equal(true, err.message.includes('Missing required argument: dependent'))
```

rodion-arr: thanks, applied

PR #32, opened 2020-08-12
cli: add more verbose descriptions to the CLI options

AshCripps:
waiting on #27 to land to add a `USAGE.md` which can then be updated here.


dominykas (approved):
Awesome!

Would be good to have a script which generates `USAGE.md`?

PR #33, opened 2020-08-27
Add negative test cases

rodion-arr:
- Add more tests for cases that throws exceptions from commands
- Add skip parameter for existing tests that do real request in case no GitHub token found (according to https://github.com/pkgjs/wiby/pull/38#issuecomment-688264931)

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -48,4 +48,17 @@ tap.test('result command', async (tap) => {
     tap.equal(true, result.includes("[ 'fake_run', 'queued' ]"))
     tap.equal(true, result.includes("[ 'fake_run_2', 'fake_conclusion' ]"))
   })
+
+  tap.test('result command handles empty response from github.getChecks()', tap => {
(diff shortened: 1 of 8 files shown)


test/cli.js
```diff
@@ -48,4 +48,17 @@ tap.test('result command', async (tap) => {
     tap.equal(true, result.includes("[ 'fake_run', 'queued' ]"))
     tap.equal(true, result.includes("[ 'fake_run_2', 'fake_conclusion' ]"))
   })
+
+  tap.test('result command handles empty response from github.getChecks()', tap => {
+    const result = childProcess.execSync(`${wibyCommand} result --dependent="https://github.com/fakeOrg/fakeRepo"`, {
```
dominykas, line 53: We should reference our own `wiby-test` here, just in case.

rodion-arr: Real requests to GitHub API are mocked in this test anyway

dominykas: Yeah, I realize that - that's the "just in case" bit ;)

rodion-arr: done

dominykas (approved):

PR #59, opened 2020-10-31
Use semantic-release

dominykas:
I'm deliberately saying that the commits _should_ follow the convention and I'm deliberately not adding a linter for that (just to avoid the burden on any potential contributors).

I'll add a config to determine bump levels based off PR labels, but until then - I'm sure we can manage without enforcing anything, so the goal here is that I can publish this using `npx semantic-release`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,35 +2,8 @@
 
 ## Code of Conduct
 
-The Node.js project has a
-[Code of Conduct](https://github.com/nodejs/admin/blob/master/CODE_OF_CONDUCT.md)
-to which all contributors must adhere.
+The Node.js Code of Conduct, which applies to this project, can be found at https://github.com/nodejs/admin/blob/master/CODE_OF_CONDUCT.md.
 
-See [details on our policy on Code of Conduct](https://github.com/nodejs/node/blob/master/doc/guides/contributing/code-of-conduct.md).
+## Commit guidelines
 
-<a id="developers-certificate-of-origin"></a>
-## Developer's Certificate of Origin 1.1
-
-By making a contribution to this project, I certify that:
-
-* (a) The contribution was created in whole or in part by me and I
-  have the right to submit it under the open source license
-  indicated in the file; or
-
-* (b) The contribution is based upon previous work that, to the best
-  of my knowledge, is covered under an appropriate open source
-  license and I have the right under that license to submit that
-  work with modifications, whether created in whole or in part
-  by me, under the same open source license (unless I am
-  permitted to submit under a different license), as indicated
-  in the file; or
-
-* (c) The contribution was provided directly to me by some other
-  person who certified (a), (b) or (c) and I have not modified
-  it.
-
-* (d) I understand and agree that this project and the contribution
-  are public and that a record of the contribution (including all
-  personal information I submit with it, including my sign-off) is
-  maintained indefinitely and may be redistributed consistent with
-  this project or the open source license(s) involved.
+- This repository uses `semantic-release` and commits should follow the [Angular Commit Message Conventions](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,4 +6,4 @@ The Node.js Code of Conduct, which applies to this project, can be found at http
 
 ## Commit guidelines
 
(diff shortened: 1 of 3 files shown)


rodion-arr (approved):
Ok for me 👍🏻

README.md
```diff
@@ -35,3 +35,8 @@ Example:
   [wiby test](./USAGE.md#wiby-test)    Test your dependents
 
   [wiby result](./USAGE.md#wiby-result) Fetch the results of your tests
+
+## Development
+
+- This repository uses `semantic-release` and commits should follow the [Angular Commit Message Conventions](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines).
```
ljharb, line 41: I’m opposed to two things here: the word “chore” in commits, and automatically publishing. Your OP implies you’re only going to manually publish, so that’s good, but i really don’t like this enforced convention, and i don’t believe semantic-release allows it to be configurable.

dominykas: Semantic release does allow it to be configurable, but I don't think it matters, because I don't think it's worth enforcing.

At work we've had massive success with a plugin which determines the bump level off a label on PR, and that has been without any controversy as well (unlike the commit message enforcement, which was a no go from day one) - I just don't have the time to configure that yet.

dominykas: I can try to reword this to avoid even the "should"?

ljharb: Only the releaser should determine the semver bump - PR reviewers/mergers aren't and shouldn't be the last line of defense. A PR label is more helpful than a commit convention, to be sure.

I like having commit conventions, but I don't think Angular's should be mentioned at all.

dominykas: I only mention the Angular convention cause it's the default and with the current volume of contributions not configuring a custom config saves time.

dominykas: Can I merge this?

ljharb: Is a custom config even possible? If we're not planning on enforcing it, then why mention it?

dominykas: Yes, [it is possible](https://github.com/semantic-release/commit-analyzer/#configuration). It doesn't even have to be driven off commit messages (I intend to configure PR labels, which is a bit more versatile).

Mentioning so that the current state is actually documented 🤷‍♂️

ljharb: Hmm, seems like we could make a package, and `config` with it relatively easily, and then that advice could go here? (assuming we agree on that package's contents, ofc)

dominykas: Yup, that's the plan, just that right now it's a bit lower on my priority list, therefore this.

ljharb: I'm not sure I understand the point of using semantic-release if you're not enforcing a commit convention, and if you're manually publishing. Can you help me understand?

dominykas: I'm not enforcing it to any drive-by contributors, but I will be following it myself, and I hope others will also follow it, but if they don't - the person making the release can push an empty commit with a message and that's that.

Determining the next bump level is only a part of `semantic-release` - it also creates release notes, posts comments on PRs which are included in the release, etc.

ljharb: tbh I'd prefer not mentioning unenforced commit conventions at all.

dominykas: Removed the mention - we can revisit later.

CONTRIBUTING.md
```diff
@@ -2,35 +2,8 @@
...
-* (c) The contribution was provided directly to me by some other
-  person who certified (a), (b) or (c) and I have not modified
-  it.
-
-* (d) I understand and agree that this project and the contribution
-  are public and that a record of the contribution (including all
-  personal information I submit with it, including my sign-off) is
-  maintained indefinitely and may be redistributed consistent with
-  this project or the open source license(s) involved.
+- This repository uses `semantic-release` and commits should follow the [Angular Commit Message Conventions](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines).
```
ljharb, line 9: this one too?

dominykas: Ooops 🤦  https://github.com/pkgjs/wiby/pull/59/commits/425030c8907c8aca1c3f5236617cf35e914e83e3

ghinks (approved):

dominykas:
Will merge in a few days - I think the comments aren't blocking?

ljharb (approved):
Still not a fan of semantic-release and non-humans publishing (nor removing 2FA protections to do it), but all my other concerns have been addressed.

PR #61, opened 2020-11-30
Switch Travis to GH Actions

dominykas:


ghinks (approved):

helio-frota (approved):

rodion-arr (approved):

BethGriggs (approved):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,44 @@
+name: Node.js CI
+
+on:
+  schedule:
+    - cron:  '23 4 * * *' # Once, nightly
+  push:
+    branches:
+      - main
+      - master
```
ljharb, line 9: why do we have both?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+
+  test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [ 15, 14, 12, 10 ]
```
ljharb, line 20: “engines” isn’t set in package.json. Is this the range it should be set to?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
```
ljharb, line 25: i have an action that encapsulates the last three steps, if we’re interested: https://github.com/ljharb/object.assign/blob/8f4cda901aff48de61267c75806d7f05bc2ca2c2/.github/workflows/node-4%2B.yml#L27

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [ 14 ]
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
```
ljharb, line 40: same here

.github/workflows/ci.yaml
dominykas: Just so that it works as soon as we switch over, without having to make another PR.

.github/workflows/ci.yaml
dominykas: I think I'd rather set the support policy, but in general the `engines` field is not really expressive enough and is probably outside of scope here anyways.

.github/workflows/ci.yaml
dominykas: It does have the benefit of making it much clearer where the matrix applies, alright, but I'm not confident about what I do or don't like about various different GH action approaches, so I'm not sure I'm ready to make this call just yet.

.github/workflows/ci.yaml
ljharb: If the `engines` field is omitted, its API is "whatever engines the `.0.0` works on", so if we don't want to accidentally support v9 or v8 or something, we'd better provide it :-)

dominykas: I did create https://github.com/pkgjs/wiby/issues/62 yesterday.

It's not like someone will come and complain that this doesn't work in 8. And if they do - I'm sure we can deal with it.

ljharb (approved):
None of my above comment are blockers.

PR #65, opened 2020-12-13
Update package.json

### LINKED ISSUE #62: Add support info
Ref: https://github.com/nodejs/package-maintenance/issues/413

Should also set the `engines` field.

rodion-arr:
- Adds support info
- Adds `engines` field

Closes #62


rodion-arr:
I'm not fully confident in `engines` field, so if I've put not correct value, will be happy to update.

Based on CI - we run on LTS + v15. But I'm not sure how `engine` field like `"^15 || ^14 || ^13 || ^12 || ^10"` will behave in case this package will be installed on non-LTS, so I just put v10+.

ljharb:
It should behave fine. If we don’t want to support node 11 (or 13, which also isn’t lts) we need the ^ form.

rodion-arr:
Thanks, updated

ljharb (approved):

BethGriggs (approved):

dominykas:
Thanks for this PR!

I'm not sure we should "support" v15 - we should use `lts_active` policy instead and remove the v15 from the engines. Testing in v15 is just to get a heads up before the next LTS.

package-support.json
```diff
@@ -0,0 +1,22 @@
+{
+  "versions": [
+    {
+      "version": "*",
+      "target": {
+        "node": "active"
```
dominykas, line 6:
```suggestion
        "node": "lts_active"
```

dominykas (approved):

PR #68, opened 2020-12-30
fix: default branch assumes *master*

ghinks:
During testing using a [parent repo](https://github.com/ghinks/my-parent) that [had a dependency upon](https://github.com/ghinks/my-dependency-id-a).

The repos did not use *master* but *main* I found issues as the default branch was hardcoded to *master*.

I created a getDefaultBranch query to get the default branch to github.js

I added both unit and integration tests to cover this.

This is the first time I have used *tap* and am glad of advice :)

Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.


ghinks:
Looking at the tests again as they pass locally.

lib/github.js
```diff
@@ -10,12 +10,42 @@ const graphqlWithAuth = graphql.defaults({
...
+      return resp.repository.defaultBranchRef.name
+    }
+    /* I'm not sure I want to return a string if the query does not
+    produce the json expected however I want to resolve this during
+    the review stage as changing this and throwing impacts other tests
+     */
+    return 'master'
+  } catch (err) {
+    if (err.errors && err.errors[0].type === 'NOT_FOUND') {
+      throw Error(`Could not find GitHub repository at https://www.github.com/${owner}/${repo}`)
```
ljharb, line 36:
```suggestion
      throw Error(`Could not find GitHub repository at https://github.com/${owner}/${repo}`)
```

lib/github.js
```diff
@@ -10,12 +10,42 @@ const graphqlWithAuth = graphql.defaults({
...
+      repo: repo
+    })
+    if (resp && resp.repository && resp.repository.defaultBranchRef) {
+      return resp.repository.defaultBranchRef.name
+    }
+    /* I'm not sure I want to return a string if the query does not
+    produce the json expected however I want to resolve this during
+    the review stage as changing this and throwing impacts other tests
+     */
+    return 'master'
```
ljharb, line 33: the default should probably be "main" since that's the default now for new github repos

lib/github.js
ghinks: agreed

lib/github.js
ghinks: agreed, and throwing an error if the branch ref cannot be found now.

ghinks:
still in progress, a few more bits to tidy up yet.

ghinks:
still a work in progress, looking at things I broke on the result call now.

lib/github.js
```diff
@@ -10,12 +10,38 @@ const graphqlWithAuth = graphql.defaults({
...
+        repository(name: $repo, owner: $owner) {
+          defaultBranchRef {
+            name
+        }
+      }
+    }`,
+      owner: owner,
+      repo: repo
+    })
+    if (resp && resp.repository && resp.repository.defaultBranchRef) {
```
dominykas, line 26: Why do we need this check? Does GH API not provide this guarantee?

lib/github.js
```diff
@@ -10,12 +10,38 @@ const graphqlWithAuth = graphql.defaults({
...
+  }
+}
+
+module.exports.getPackageJson = async function getPackageJson (owner, repo, defaultBranch) {
   try {
     const resp = await graphqlWithAuth({
       query: `query($owner: String!, $repo: String!) {
         repository(name: $repo, owner: $owner) {
-          object(expression: "master:package.json") {
+          object(expression: "${defaultBranch}:package.json") {
```
dominykas, line 44:
```suggestion
          object(expression: "HEAD:package.json") {
```

This should read the `package.json` directly off the default branch, allowing us to skip that extra query?

ghinks: Maybe, let me look at this too. I have found the API from GIthub a bit hard to read and if this is better then we should do the simple thing.

ghinks:
I had a  look at the query you suggested and it looks like a much better way to do things. I will close this PR and open another with that change.

Issue #56, opened 2020-10-27
Example usage with Github Actions


dominykas:
- Something to kick off `wiby test` (triggered off a comment?)
- Something to poll for results and report them back as a comment

dominykas:
OK, I think this issue is on the critical path - the CLI approach is all nice, but `wiby` needs to happen in CI.

I tentatively opened this referencing Github Actions, but actions don't make this easy, so I'll need to do some digging.

Kicking off the initial `wiby test` is the easy part.

The hard part is to collect the results, because you can't trigger an action off a status change in another repo. Which means some sort of polling, but there's no obvious way of implementing that using actions either, or at least no obvious way which wouldn't use up the precious actions minutes.

There could be a scheduled which basically polls all the things that are in flight, but I'm struggling to imagine a way to make it easy to set up, as GH actions don't have a way of sharing workflows.

Neither is this easy on Travis - you'd essentially have to keep a job running until it times out or the dependents finish their jobs, which can be rather wasteful as well.

I'll ponder other available tools to achieve this, but ideas are most welcome!

dominykas:
OK, so here's the plan:

1. For the initial version, there will not be an automated results check. I don't have a good solution to save on minutes. So you will have to use a comment to `wiby test`, which will kick off the tests, and it will also set a status check to pending. You will then have to explicitly comment to `wiby result`, which will loop through the tests that were kicked off and update the status check accordingly.
    - I will open an issue on GH side to add a "sleep" step (which wouldn't count against minutes) or to add a "delayed trigger" (which would explicitly kick off X minutes after Y event or X minutes after an explicit API call).
2. There will be several GH workflows involved, so I'll add a new command to sync up the canonical workflows. Not sure about the combo of commands - an interactive `wiby install` would be awesome, but this also needs a silent way to just update the options. Maybe a `wiby outdated` and `wiby update` make sense. I'm thinking to do this as part of `wiby`, but it probably makes sense to extract as a separate package as well, if the approach proves itself viable.

dominykas:
Things not covered in #69, to be reconsidered if we want them for MVP:

- It seems the the `details_url` does not do what I want it to do... Basically, the status check under the PR links to the details about the status check, but somehow does not want to link to the actual build details.
[...]

PR #69, opened 2020-12-31
Github Workflow

dominykas:
See https://github.com/pkgjs/wiby/pull/69/files#diff-00505182d7d4129a98ee7fb7b652a1ef6b608d7620a681914db3a7a4903eee9f for usage details.

Marking this ready for review to get feedback, but we can't merge this without the following:

- [x] Create a bot account - https://github.com/wiby-bot
- [x] Add the `WIBY_TOKEN`

I have also added a bunch of notes on https://github.com/pkgjs/wiby/issues/56#issuecomment-768301949 on how to move this forward, but I figure I'll want to make some changes to the wiby CLI first, rather than the GH actions, and I'd also like to start getting this to a more demoable place and also to start test driving this on wiby itself, even if it still requires some manual interaction (namely deleting the `wiby-wiby` branch in the `wiby-test` org).

### DIFF
--- a/docs/github-workflow.md
+++ b/docs/github-workflow.md
@@ -1,0 +1,18 @@
+# wiby github workflow
+
+The `wiby` workflow for Github is very much work in progress, but please ask questions in the issue tracker.
+
+## Installation / Upgrading
+
+This is a manual step, until automation is ready:
+
+1. Install `wiby` as a dev dependency (and update to latest version)
+   - having it as a dev dep is not strictly necessary, as the workflow relies on `npx wiby`, but it does give control over the installed version
+2. `mkdir -p .github/workflows && npm explore wiby -- cat .github/workflows/wiby.yaml > .github/workflows/wiby.yaml` 
+   - Duplicate the default `wiby` workflow - the automation, when developed, will attempt to keep it up to date at that location
+3. Make sure your `.github/workflows/wiby.yaml` ends up on the default (`main`, `master`) branch on Github
+   - i.e. commit, push, open a PR and merge it, if necessary
+4. Add a `WIBY_TOKEN` secret
+   - create a new one at https://github.com/settings/tokens, required scopes: `repo:status` and `public_repo`
+   - the account has to have push access to the test repositories - you will probably want to fork the dependents into that account (or a separate organization)
+   - you may want to use a robot account which _does not have **any** merge access_, as Github does not allow restricting tokens to a subset of repositories
--- a/docs/github-workflow.md
+++ b/docs/github-workflow.md
@@ -14,8 +14,9 @@ This is a manual step, until automation is ready:
    - i.e. commit, push, open a PR and merge it, if necessary
 4. Add a `WIBY_TOKEN` secret
    - create a new one at https://github.com/settings/tokens, required scopes: `repo:status` and `public_repo`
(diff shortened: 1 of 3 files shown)


ghinks:
I'm going to review to the best of my abilities here. The instructions are clear and I could follow and understand them. I am still getting up to speed on github actions and am learning from reading the wiby.yaml file. So will not review that section.

dominykas:
A "wiby test" comment in a PR, and then a "wiby result" comment to check for results. Less, then ideal :( but best I could come up with (and build in short time to get a PoC).

ghinks:
I carried out a bit of experimentation by taking the changes and releasing a wiby module in a private namespace. When updating a dependency using the instructions in the PR the wiby action does the push to the dependent repo but does not get marked as completed.

Screen Shot 2021-01-09 at 11 30 11 AM

dominykas:
If you comment "wiby result" _after_ all the tests completed - it should go green.

Do you have a link to the PR? I can take a look at the output...

ghinks:
yes, ( apologies , work got in the way ).

[PR dependency](https://github.com/ghinks/my-dependency-id-a/pull/5#issuecomment-755843009)

ghinks:
Screen Shot 2021-01-12 at 3 06 11 PM

did as suggested ( still getting up to speed on actions ) , seemed to do the trick.

dominykas:
Thanks for the reviews and test here @ghinks 🙇

docs/github-workflow.md
```diff
@@ -16,3 +16,15 @@ This is a manual step, until automation is ready:
    - create a new one at https://github.com/settings/tokens, required scopes: `repo:status` and `public_repo`
    - the account has to have push access to the test repositories - you will probably want to fork the dependents into that account (or a separate organization)
    - you may want to use a robot account which _does not have **any** merge access_, as Github does not allow restricting tokens to a subset of repositories
+
+
+## Usage
+
+0. This assumes `.wiby.json` is already configured and that you're an owner or a contributor on the dependency repo
+1. Open a PR against the dependency repository
+2. Post a comment saying `wiby test` in the PR
```
ljharb, line 26: Maybe this should be a label instead? It’s cleaner and only permissioned people can do it.

dominykas: The issue with the label is that it's unclear how to re-run the job. Remove the label and add it again? Automatically remove the label on new push or when the job succeeds?

Happy to iterate on this in the upcoming versions - would like to get this into some real life usage though.

ljharb: Starting the job would remove the label, so you'd just "add the label".

I also think that posting the results should be automatic, in a single updated-in-place comment, rather than requested.

dominykas: It should be automatic, yes, but there is no simple way to achieve it with what is available. Ideally you'd be able to schedule a job or to have a sleep routine which doesn't count against your minutes, but short of that - there's a significant amount of work needed, hence why I started with a comment.

ljharb: ah, gotcha. (either way this is fine for now, just offering thoughts)

dominykas: Thanks, ideas most welcome.

dominykas:
@pkgjs/wiby-collaborators can I get some ✅ here please?

rodion-arr (approved):

PR #77, opened 2021-01-27
Fix up integration tests

dominykas:
- `wiby-test` org uses the real package name (i.e. `wiby`) - not `@pkgjs/wiby`, which we use when unit testing.
- cleans up some tests so that they don't break after https://github.com/wiby-test/pass/pull/1 / https://github.com/wiby-test/fail/pull/1 / https://github.com/wiby-test/partial/pull/1 are merged (can I get some 👍 there please?)
- adds the `GITHUB_TOKEN` so that we can get integration tests during CI runs (although I'd like to retire most of them and replace with true integration tests, i.e. avoid exporting functions for the sake of testing)


test/test-integration.js
```diff
@@ -1,12 +0,0 @@
-require('dotenv').config()
-const tap = require('tap')
-const CONFIG = require('./fixtures/config')
-const pkgTest = require('../lib/test')
-
-tap.test('Test correct sha returned for a GitHub repository', async tap => {
```
dominykas, line 6: This removal is mostly triggered by the upcoming changes in the wiby-test/pass repo, which would require us to update `DEP_HEAD_SHA`. Maintaining this is a low value activity, since we already have unit test coverage for this and we'll soon be able to test integration via the wiby action.

lib/test.js
```diff
@@ -32,7 +32,7 @@ module.exports = async function ({ dependents }) {
   }
 }
 
-const getCommitHash = module.exports.getCommitHash = async function getCommitHash (owner, repo) {
```
dominykas, line 35: Technically, this is a breaking change, but the exports removed were only ever exposed for testing purposes and were never intended to be used as an API. We're pre-v1 anyways.

rodion-arr (approved):

ljharb (approved):

PR #81, opened 2021-02-22
`wiby clean`

### LINKED ISSUE #79: `wiby clean` command
At the moment, `wiby test` creates a branch called `wiby-[dep name]` in the dependent repos. There is nothing to clean this up - running a new `wiby test` will fail if the branch already exists.

`wiby test` should not fail on existing branch - but that is a separate, larger problem (I'll create a separate issue), but in the meantime - deleting the `wiby-wiby` branch manually on `wiby-test/*` repos is a hassle during development.

`wiby clean` should take the following options:

- `--all` - remove all `wiby-*` branches (by default only remove the current branch, based on the naming as proposed in #78)
- `--dry-run` - show the branches that would be removed
    - alternative: do the dry-run by default and ask the user to confirm the action? With an optional `--force` to skip the confirmation?
    - bonus points: what if the branches contain commits other than ones made by wiby?

dominykas:
Closes #79

### DIFF
--- a/lib/clean.js
+++ b/lib/clean.js
@@ -5,12 +5,12 @@ const gitURLParse = require('git-url-parse')
 const logger = require('./logger')
 
 // setup logger namespace
-const testCommandNamespace = 'wiby:clean'
-const debug = logger(testCommandNamespace)
+const cleanCommandNamespace = 'wiby:clean'
+const debug = logger(cleanCommandNamespace)
 
 module.exports = async ({ dependents }, { all, dryRun }) => {
-  // enable log output for test command without DEBUG env
-  logger.enableLogs(testCommandNamespace)
+  // enable log output for clean command without DEBUG env
+  logger.enableLogs(cleanCommandNamespace)
 
   const parentPkgJSON = await require('./test').getLocalPackageJSON()
   const parentPkgInfo = gitURLParse(parentPkgJSON.repository.url)
--- a/lib/clean.js
+++ b/lib/clean.js
@@ -1,0 +1,34 @@
+'use strict'
+
+const github = require('./github')
+const gitURLParse = require('git-url-parse')
+const logger = require('./logger')
+
+// setup logger namespace
+const testCommandNamespace = 'wiby:clean'
+const debug = logger(testCommandNamespace)
+
+module.exports = async ({ dependents }, { all, dryRun }) => {
+  // enable log output for test command without DEBUG env
+  logger.enableLogs(testCommandNamespace)
+
+  const parentPkgJSON = await require('./test').getLocalPackageJSON()
+  const parentPkgInfo = gitURLParse(parentPkgJSON.repository.url)
+  debug(`Parent module: ${parentPkgInfo.owner}/${parentPkgJSON.name}`)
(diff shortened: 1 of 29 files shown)


lib/clean.js
```diff
@@ -0,0 +1,44 @@
...
+const github = require('./github')
+const gitURLParse = require('git-url-parse')
+const logger = require('./logger')
+
+// setup logger namespace
+const testCommandNamespace = 'wiby:clean'
+const debug = logger(testCommandNamespace)
+
+module.exports = async ({ dependents }, { all, dryRun }) => {
+  // enable log output for test command without DEBUG env
```
rodion-arr, line 12:
```suggestion
  // enable log output for clean command without DEBUG env
```

lib/clean.js
```diff
@@ -0,0 +1,44 @@
+'use strict'
+
+const github = require('./github')
+const gitURLParse = require('git-url-parse')
+const logger = require('./logger')
+
+// setup logger namespace
+const testCommandNamespace = 'wiby:clean'
```
rodion-arr, line 8:
```suggestion
const cleanCommandNamespace = 'wiby:clean'
```

lib/clean.js
```diff
@@ -0,0 +1,44 @@
...
+const debug = logger(testCommandNamespace)
```
rodion-arr, line 9:
```suggestion
const debug = logger(cleanCommandNamespace)
```

test/clean.js
```diff
@@ -0,0 +1,36 @@
...
+  nock.enableNetConnect()
+})
+
+tap.test('wiby.clean()', async (tap) => {
+  tap.test('should check if the wiby branch exists', async (tap) => {
+    nock('https://api.github.com')
+      .get(`/repos/wiby-test/${CONFIG.DEP_REPO}/branches/wiby-wiby`)
+      .reply(404)
+
+    await wiby.clean({ dependents: [{ repository: `https://www.github.com/${CONFIG.DEP_ORG}/${CONFIG.DEP_REPO}` }] }, {})
```
rodion-arr, line 24: Could you please elaborate what is assertion here? It shouldn't throw exception?

rodion-arr (approved):
👍🏻

lib/clean.js
dominykas: https://github.com/pkgjs/wiby/pull/81/commits/0108df89afe72267c043b8e6e18d56de39c17584

lib/clean.js
dominykas: https://github.com/pkgjs/wiby/pull/81/commits/0108df89afe72267c043b8e6e18d56de39c17584

lib/clean.js
dominykas: https://github.com/pkgjs/wiby/pull/81/commits/0108df89afe72267c043b8e6e18d56de39c17584

test/clean.js
dominykas: https://github.com/pkgjs/wiby/pull/81/commits/196d30a1d94ea402a658ab91844d1131d45044b9

PR #82, opened 2021-02-23
Cleanup internally shared functions and rearrange tests folder

dominykas:
This is just to make things a teeny bit tidier (primarily for myself).


ghinks (approved):

lib/test.js
```diff
@@ -25,7 +25,7 @@ module.exports = async function ({ dependents }) {
     const dependentPkgJSON = await github.getPackageJson(dependentPkgInfo.owner, dependentPkgInfo.name)
     debug(`Dependent module: ${dependentPkgInfo.owner}/${dependentPkgInfo.name}`)
 
-    if (!checkPackageInPackageJSON(parentPkgJSON.name, dependentPkgJSON)) {
+    if (!context.checkPackageInPackageJSON(parentPkgJSON.name, dependentPkgJSON)) {
```
ghinks, line 28: **my only comment so far**

The call to **applyPatch**, which takes place immediately after this check, carries out the same check 
```javascript
  if (!context.checkPackageInPackageJSON(module, packageJSON)) {}
```

but with less debug info, maybe just move this extra debug for errors into **applyPatch**.

dominykas: Sounds good - but since we both have a PR open in the similar area - mind if we merge this and then maybe you can do it in yours or maybe we can follow this up with a separate PR after yours is mered?

PR #84, opened 2021-03-02
Feature/other/dependencies

ghinks:
handle 

- devDependencies
- peerDependencies
- optionalDependencies

in the dependent package. 

Changes were made on top [cleanup branch](https://github.com/dominykas/wiby/tree/cleanup) this PR is to provide visibility of the changes.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -13,7 +13,7 @@ exports.getLocalPackageJSON = async function getLocalPackageJSON () {
 
 /*
 The check should search both the dev and peer dependencies to find out if the dependency is
-regular, dev or peer. We only want to check on this the once :)
+regular, dev or peer.
  */
 exports.checkPackageInPackageJSON = function checkPackageInPackageJSON (dep, packageJSON) {
   const dependencyKeyNames = [
--- a/lib/context.js
+++ b/lib/context.js
@@ -23,8 +23,5 @@ exports.checkPackageInPackageJSON = function checkPackageInPackageJSON (dep, pac
     'optionalDependencies'
   ]
   const checkKeyHasDep = (key) => Object.prototype.hasOwnProperty.call(packageJSON, key) && Object.prototype.hasOwnProperty.call(packageJSON[key], dep)
-  for (let i = 0; i < dependencyKeyNames.length; i++) {
-    if (checkKeyHasDep(dependencyKeyNames[i])) return dependencyKeyNames[i]
-  }
-  return false
+  return dependencyKeyNames.find(x => checkKeyHasDep(x))
 }
--- a/lib/context.js
+++ b/lib/context.js
@@ -11,6 +11,20 @@ exports.getLocalPackageJSON = async function getLocalPackageJSON () {
   return JSON.parse(pkg)
 }
 
+/*
+The check should search both the dev and peer dependencies to find out if the dependency is
+regular, dev or peer. We only want to check on this the once :)
+ */
 exports.checkPackageInPackageJSON = function checkPackageInPackageJSON (dep, packageJSON) {
-  return Object.prototype.hasOwnProperty.call(packageJSON.dependencies, dep)
+  return (
+    (Object.prototype.hasOwnProperty.call(packageJSON, 'dependencies') && Object.prototype.hasOwnProperty.call(packageJSON.dependencies, dep)
+      ? 'dependencies'
+      : false) ||
+    (Object.prototype.hasOwnProperty.call(packageJSON, 'devDependencies') && Object.prototype.hasOwnProperty.call(packageJSON.devDependencies, dep)
+      ? 'devDependencies'
+      : false) ||
+    (Object.prototype.hasOwnProperty.call(packageJSON, 'peerDependencies') && Object.prototype.hasOwnProperty.call(packageJSON.peerDependencies, dep)
+      ? 'peerDependencies'
+      : false)
+  )
 }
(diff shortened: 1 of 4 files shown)


lib/context.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+exports.checkPackageInPackageJSON = function checkPackageInPackageJSON (dep, packageJSON) {
+  const dependencyKeyNames = [
+    'dependencies',
+    'devDependencies',
+    'peerDependencies',
+    'optionalDependencies'
+  ]
+  const checkKeyHasDep = (key) => Object.prototype.hasOwnProperty.call(packageJSON, key) && Object.prototype.hasOwnProperty.call(packageJSON[key], dep)
+  for (let i = 0; i < dependencyKeyNames.length; i++) {
```
rodion-arr, line 26: Looks like `for..of` loop will look here more simple?

lib/context.js
```diff
@@ -0,0 +1,30 @@
...
+    if (checkKeyHasDep(dependencyKeyNames[i])) return dependencyKeyNames[i]
+  }
+  return false
```
ljharb, line 29:
```suggestion
  return dependencyKeyNames.find(x => checkKeyHasDep(x))
```
loops are rarely the simpler approach :-)

ghinks: thank you for the reviews, update made.

test/fixtures/config.js
```diff
@@ -8,7 +8,33 @@ module.exports.PATCHED_PKGJSON = Object.assign({}, module.exports.PKGJSON, {
...
   )
 })
+
+module.exports.PKGJSON_DEV_DEPS = Object.assign({}, module.exports.PKGJSON, {
+  devDependencies: {
+    '@other/name': '*'
+  }
+})
+module.exports.PATCHED_DEV_DEPS_PKGJSON = Object.assign({}, module.exports.PKGJSON, {
+  devDependencies: Object.assign(
```
dominykas, line 18: Is there a need for this extra `Object.assign`?

test/fixtures/config.js
```diff
@@ -8,7 +8,33 @@ module.exports.PATCHED_PKGJSON = Object.assign({}, module.exports.PKGJSON, {
...
+  )
+})
+
+module.exports.PKGJSON_PEER_DEPS = Object.assign({}, module.exports.PKGJSON, {
+  peerDependencies: {
+    '@other/plugin': '*'
+  }
+})
+module.exports.PATCHED_PEER_DEPS_PKGJSON = Object.assign({}, module.exports.PKGJSON, {
+  peerDependencies: Object.assign(
```
dominykas, line 30: Is there a need for this extra `Object.assign`?

test/fixtures/config.js
```diff
@@ -8,7 +8,33 @@ module.exports.PATCHED_PKGJSON = Object.assign({}, module.exports.PKGJSON, {
     { '@pkgjs/wiby': 'pkgjs/wiby#577c08e8fd5e1b3156ce75b2e5d9e3023dac180e' }
   )
 })
+
+module.exports.PKGJSON_DEV_DEPS = Object.assign({}, module.exports.PKGJSON, {
+  devDependencies: {
+    '@other/name': '*'
```
dominykas, line 14: Do we need to use a different package name here? If we do - we should maybe stick to `@wiby-test/*` and `@pkgjs/wiby`, just in case.

dominykas (approved):
Minor nits, otherwise - 👍

ghinks:
re-merged main back into this branch to make diff clear

test/fixtures/config.js
ghinks: I think the test is actually best served with this as it is.

This is my reasoning we need to make sure that we can find the "@other/name" patch when that is the *dev* dependency in question even when "@pkgjs/wiby" is a regular dependency.

```json
{
  "name": "@wiby-test/pass",
  "private": true,
  "version": "1.0.0",
  "description": "A dependent of `wiby` which always passes the tests",
  "scripts": {
    "test": "true"
  },
  "repository": {
    "type": "git",
    "url": "git+https://github.com/wiby-test/pass.git"
  },
  "license": "MIT",
  "dependencies": {
    "@pkgjs/wiby": "*"
  },
  "devDependencies": {
    "@other/name": "other/name#577c08e8fd5e1b3156ce75b2e5d9e3023dac180e"
  }
}
```

The alternate is to create another passing package.json fixture with an empty dependencies and wiby in the devDependencies. 

Let me create another PR for that as you have passed this already.

test/fixtures/config.js
ghinks: addressed.

test/fixtures/config.js
ghinks: addressed.

ghinks:
@rodion-arr , another re-review please.

lib/context.js
```diff
@@ -11,6 +11,17 @@ exports.getLocalPackageJSON = async function getLocalPackageJSON () {
   return JSON.parse(pkg)
 }
 
+/*
+The check should search both the dev and peer dependencies to find out if the dependency is
+regular, dev or peer. We only want to check on this the once :)
```
rodion-arr, line 16: Not sure if `the` needed here before `once`?

rodion-arr (approved):

lib/context.js
ghinks: agreed and updated accordingly.

test/fixtures/config.js
dominykas: That's fine, we do need this test, I was just thinking that we should use e.g. `@wiby-test/other-name`, rather than `@other/name`, because `@other` may exist and is not owned by us, so if we make a mistake (or someone uses some automated scanner or smth), we might be sending traffic the wrong way. Unlikely, but still.

Either way, I'll merge this. Not a blocker.

Issue #78, opened 2021-01-31
Name wiby test branches based on the dependency branch name


dominykas:
At the moment, `wiby test` will create a branch called `wiby-[dependency name]` on the dependent being tested (and `wiby result` will check here).

This is very restricting - it means that you can only test one set of changes in the dependency at a time (i.e. if you have multiple PRs open on `wiby`, they would both try to write into `wiby-wiby`). Moreover, this does not allow automated linking of multiple dependencies into a single test branch.

The proposal here is to name the branch based on the dependency branch, i.e. if I'm testing `feat/new-stuff` on `wiby`, then the test branch would be called `wiby-feat/new-stuff` on the dependent. This way other dependencies could have `feat/new-stuff` and would all link into the same `wiby-feat/new-stuff` branch.

- The `wiby-` prefix is necessary, to avoid touching non-wiby branches that people may be working on. 
- If the dependency branch already starts with `wiby-` - the prefix will not get added multiple times. Rationale: people may need to make tweaks to make the tests pass, esp. when linking multiple dependencies together.
- If the dependent already has a branch called `feat/new-stuff`, then `wiby` should base the `wiby-feat/new-stuff` off that, rather than off `main` / `master`.

(I'll add some examples shortly, to clarify this).

PR #85, opened 2021-03-18
Name wiby test branches based on the dependency branch name

### LINKED ISSUE #83: Link the current branch/commit, rather than default branch
At the moment, `wiby test` will link the `sha` of the `master` into the dependents.

It should instead link the branch that it's running against. Possibly try to detect the branch using `env-ci` if it's running on a detached head.

I'll fix this together with #78 - just documenting the issue.

dominykas:
- Implements a large chunk of #78 
- Closes #83 
- You probably want to ignore white-space when reviewing: https://github.com/pkgjs/wiby/pull/85/files?w=1

### DIFF
--- a/lib/clean.js
+++ b/lib/clean.js
@@ -24,7 +24,8 @@ module.exports = async ({ dependents }, { all, dryRun }) => {
 
     let branches
 
-    const branch = await context.getTestingBranchName(parentPkgJSON.name)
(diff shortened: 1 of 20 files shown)


dominykas:
wiby test

rodion-arr (approved):

PR #86, opened 2021-03-25
Add github-workflow install scripts

dominykas:
This might prove to be useful even when github starts providing us with better primitives for sharing entire workflows.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -22,6 +22,27 @@ Options:
 ```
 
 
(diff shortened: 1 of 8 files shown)


dominykas:
wiby test

dominykas:
wiby test

dominykas:
wiby test

rodion-arr (approved):

PR #91, opened 2021-03-31
Upgrade `tap` to v15

rodion-arr:
We were affected by 2 major changes in `tap` v15:

1. Separate `tap.has` (`tap.include()`) from `tap.match`, so these are distinct.
It was required to change `tap.include()` to `tap.match()` as next code is behaves on v15 like this:
```javascript
tap.test('test 15', async (tap) => {
  const string = 'some text expected'
  tap.include(string, 'expected') // will fail
  tap.match(string, 'expected') // will pass
})
````
2. `--check-coverage` on by default.
Tap started to fail by default in case coverage is under 100%.
So I added few more tests to be in "100% Coverage Club 🏅"

Related to #89

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   "devDependencies": {
     "nock": "^13.0.3",
     "standard": "^16.0.0",
-    "tap": "^14.10.7",
+    "tap": "^15.0.0",
     "tmp": "^0.2.1"
   },
   "standard": {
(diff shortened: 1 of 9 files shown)


dominykas (approved):
🌟

ghinks (approved):

ghinks:
I had a bit of trouble getting the tests to pass. Even when I pulled from Rodion's branch there were issues.

rodion-arr:
For it was required to update git version due to `git init --initial-branch` command was failing.

ghinks:
many thanks @rodion-arr , I'm good again.

BethGriggs (approved):

Issue #53, opened 2020-10-27
Open pull-requests when testing


dominykas:
Ref: https://github.com/pkgjs/wiby/blob/master/docs/DRAFT-FLOW.md step 9

> Open a draft PR in the dependents repository that contains the patched `package.json`.

Depending on the CI setup, it might not be enough to just push the branch - `wiby.json` should allow setting a configuration option to additionally open a PR.

This also means that we'd probably need to close the PR, etc, when the tests are done.

dominykas:
@ghinks and maybe this one?

ghinks:
ok, will look at this one over the next few days.

dominykas:
Let me know if you want to discuss it beforehand!

ghinks:
I think we should discuss it. Let me re-read and respond today. Been super busy.

ghinks:
Had a quick look, implementation looks straightforwards enough. However there will always be questions.

## Options

- CLI Option Name. Should we have a command line option like **pull-request**?
- Wiby config option. Should we have an option in the configuration pull-request: boolean?

There are some gotchas. 
- Just because we have a github token it does not mean it can create a PR. Do we want to check , not sure we can yet but will look into it
- Should we check that the PR has not already been raised? I think when I played with the API it would not let me raise the same PR twice even with different titles

I'll put something together in draft.

ghinks:
@dominykas @rodion-arr I was wondering if it would be a good idea to split this into 2 parts. 

- the first raising the draft PR
- the second checking and closing the PR if the tests passed

your thoughts?

I'm trying to avoid bigger PRs.

rodion-arr:
It looks like pretty much separate features, imo. I'm not against separate PRs but maybe it's better to release them together (if we have any users at all at this stage)

ghinks:
yes agreed, release together. But I shall raise 2 PRs.

dominykas:
> CLI Option Name. Should we have a command line option like pull-request?

Yeah, CLI convention is for words to be separated by `-` in the args, and Yargs will automatically `camelCase` that when using in the code.

> Wiby config option. Should we have an option in the configuration pull-request: boolean?

I suppose since Yargs transforms that into `camelCase`, the `wiby.json` should also use `pullRequest`?

> Just because we have a github token it does not mean it can create a PR. Do we want to check , not sure we can yet but will look into it

[...]

PR #93, opened 2021-04-10
Feature/close draft pr

ghinks:
Changes for [issue 53](https://github.com/pkgjs/wiby/issues/53)

Both opens the draft PR on **test** and  closes it on **result**

using --pullrequest, open to use of other names

goes in ontop of [pull 91 tap update](https://github.com/pkgjs/wiby/pull/91)


.wiby.json
```diff
@@ -9,5 +9,6 @@
     {
       "repository": "git+https://github.com/wiby-test/pass"
     }
-  ]
+  ],
+  "pullrequest": false
```
dominykas, line 13: I think this should be an option for each dependent. This does not depend on `wiby` user, but on the configuration of CI in the target repo.

lib/github.js
```diff
@@ -136,3 +136,24 @@ module.exports.getCommitStatusesForRef = async function getCommitStatusesForRef
...
+    repo,
+    title,
+    head,
+    base,
+    draft,
+    body
+  })
+}
+
+module.exports.updatePRStatus = async function updatePRStatus (owner, repo, pullNumber) {
```
dominykas, line 152: This doesn't allow you to set arbitrary state, so maybe a name change is in order?

```suggestion
module.exports.closePr = async function closePr (owner, repo, pullNumber) {
```

lib/result.js
```diff
@@ -37,6 +37,7 @@ module.exports = async function ({ dependents }) {
     const parentBranchName = await context.getParentBranchName()
     const branch = await context.getTestingBranchName(parentBranchName)
     let resp = await github.getChecks(dependentPkgInfo.owner, dependentPkgInfo.name, branch)
+    await updatePRStatus(dependentPkgInfo.owner, dependentPkgInfo.name, branch, resp.data.check_runs)
```
dominykas, line 40: 🤔

So we do have a `wiby clean` command, which deletes the branches. So maybe that's a better fit for this?

I'm kind of torn on whether the full `wiby clean` should happen automatically after the results are successful (it seems it now does it regardless of the result?) - I think if, as an implementer of a feature, I run `wiby result`, and there's a failure - I definitely want to keep things as is, so that I can go explore and debug. But if everything passed - I might still want to keep things around, if I'm building a feature that spans multiple repos?

Quite possibly this is a full-on separate feature or needs to be behind a global config flag (`--auto-clean` or smth).

.wiby.json
dominykas:
```suggestion
  "pullRequest": false
```

And `--pull-request` for params in `commands`?

Prior-art: `dry-run` in `wiby clean` and also Yargs will automagically convert the `-` to `camelCase` too.

dominykas:
👏 sorry for not responding and not clarifying things earlier... I see there's some other discussions on the issue - will check those out now 👍

.wiby.json
dominykas: Just figured this might be ambiguous, so...

1. If we're calling `wiby test --dependent [dependent url]`, then we're only running for a single dependent, and `--pull-request` applies to just that - already implied and works
2. If we're calling `wiby test --pull-request` for _all_ dependents, then the CLI option needs to override any configuration options, i.e. it should open pull requests for all dependents defined in `wiby.json`
3. If we're calling `wiby test` without `--pull-request`, then it should only open pull requests for dependents that have `pullRequest: true` configured.

Would that make sense?

ghinks: yes, this makes sense.

lib/github.js
ghinks: 👍

lib/result.js
ghinks: agreed.
Let me address the other issues in this draft PR. Then we can discuss further and maybe move it behind a different command.

lib/test.js
```diff
@@ -58,3 +61,15 @@ async function pushPatch (dependentPkgJson, dependentOwner, dependentRepo, paren
   await github.createBranch(dependentOwner, dependentRepo, commitSha, branch)
   debug(`Changes pushed to https://github.com/${dependentOwner}/${dependentRepo}/blob/${branch}/package.json`)
 }
+
+const createPR = module.exports.closePR = async function closePR (dependentOwner, dependentRepo, parentBranchName, parentDependencyLink) {
```
rodion-arr, line 65:
```suggestion
const createPR = module.exports. createPR = async function createPR (dependentOwner, dependentRepo, parentBranchName, parentDependencyLink) {
```

lib/result.js
```diff
@@ -162,3 +165,27 @@ function resolveCodeAndExit (overallStatus) {
     process.exit(PENDING_RESULT_EXIT_CODE)
   }
 }
+
+const closePR = module.exports.closePR = async function closePR (owner, repo, branch, checkRuns) {
+  if (!checkRuns || checkRuns.length === 0) {
```
dominykas, line 170: The `length` check here (and below) is probably not necessary - the array iterators will be effectively a no-op even without these checks, so there's fewer tests to write to ensure coverage.

lib/result.js
```diff
@@ -162,3 +165,27 @@ function resolveCodeAndExit (overallStatus) {
...
+    return
+  }
+  const prsToClose = checkRuns.reduce((acc, check) => {
+    if (check.status === 'completed' &&
+        check.conclusion === 'success' &&
```
dominykas, line 175: I forgot... do we not already have a helper function to verify if check has passed?

lib/test.js
```diff
@@ -58,3 +61,15 @@ async function pushPatch (dependentPkgJson, dependentOwner, dependentRepo, paren
...
+  const title = `Wiby changes to ${parentDependencyLink}`
+  const body = `Wiby has raised this PR to kick off automated jobs.
+  You are dependent upon ${parentDependencyLink}
+  `
+  const head = context.getTestingBranchName(parentBranchName)
+  const draft = true
+  const result = await github.createPR(dependentOwner, dependentRepo, title, head, parentBranchName, draft, body)
```
dominykas, line 72: Do we need to set the `base` to `parentBranchName` here? I do have a note in the issues, that if the branch exists in the dependent repo as well as in parent repo, we should try to use that, but that's not implemented yet, so at this point `parentBranchName` likely does not even exist - we should probably keep it `undefined` (if the API allows it - otherwise we need to fetch the default branch name...)?

ghinks:
I'm redoing this with the suggestion that it be a different feature flag. Will read comments this afternoon

lib/result.js
ghinks: 👍

lib/test.js
ghinks: 👍

lib/result.js
ghinks: not sure I see this else where?

lib/test.js
ghinks: I was not sure on this comment, but its late afternoon so I'm taking it easy.

ghinks: @dominykas ok, I came back to look at this comment and confess to not understanding at first you are referring to this I think [issue 95](https://github.com/pkgjs/wiby/issues/95). If so I think I understand.

dominykas:
🤔 I see you added `close-pr` as a new command. I guess it might be useful on its own. 👍 

That said, I think we probably still need to add it into `wiby clean` as well? Because `wiby clean` won't be able to delete branches if they have PRs open.

lib/result.js
dominykas: Yeah, OK, the one we have is in a different context 👍

lib/test.js
dominykas: It's kind of related to that issue, but that's not exactly it.

When opening a PR, we want to merge `head` into `base`. So most of the time `base` should be `main` / `master` (i.e. default branch), not `parentBranchName`?

0. Suppose you're working on `my-feature`
1. You `wiby test`
2. A `wiby-my-feature` branch is created in the dependent repo
3. You should be opening a PR on the dependent repo to merge `wiby-my-feature` into `main`

i.e. `my-feature` (`parentBranchName`) will not even exist in the dependent, most of the time. When it does - we do have the "existing branches" case, which needs to be thought through - most likely we'll want to create a `wiby-my-feature` in the dependent, but we will not start it at `main` / `master`, but at `my-feature` instead. In that case, there could be some DX wins if the PR on the dependent tried to merge `wiby-my-feature` into `my-feature`, but at the same time - if it tries to merge `wiby-my-feature` into `main`, it's probably no big deal either?

lib/test.js
```diff
@@ -32,6 +32,9 @@ module.exports = async function ({ dependents }) {
 
     const patchedPackageJSON = applyPatch(parentDependencyLink, parentPkgJSON.name, dependentPkgJson, parentPkgJSON.name)
     await pushPatch(patchedPackageJSON, dependentRepositoryInfo.owner, dependentRepositoryInfo.name, parentPkgJSON.name, parentBranchName)
+    if (pullRequest) {
```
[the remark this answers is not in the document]
ghinks: Ok. I agree that clean needs it to. I'll merge the latest changes back into my branch and carry on. 

One other thing occurred to me. We may want to be using the Promise.allSettled() module as if we are closing multiple PRs we should report all the failures and successes accurately?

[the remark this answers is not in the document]
dominykas: > Promise.allSettled() 

Yeah, that probably makes sense.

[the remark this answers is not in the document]
ghinks: After looking at the needs of the clean command. I do not think we need to do anything. If the branch is deleted during the clean. Then the PR that was raised against that branch to trigger an action is also deleted. 

If you agree with my opinion I'll look at [issue 95](https://github.com/pkgjs/wiby/issues/95) with respect to the close-pr command.

rodion-arr (approved):

bin/commands/close-pr.js
```diff
@@ -0,0 +1,35 @@
...
+    desc: 'URL of a dependent',
+    type: 'string',
+    conflicts: 'config'
+  })
+  .option('close-pr', {
+    desc: 'Close a PR of a dependent raised in test',
+    alias: 'pr',
+    type: 'boolean',
+    conflicts: 'config'
+  })
```
dominykas, line 18:
```suggestion
```

Doesn't seem to be used?

ghinks:
many thanks for the reviews

dominykas:
I'm keen to merge this, but https://github.com/pkgjs/wiby/pull/93#discussion_r615362448 is not yet resolved? (the `parentBranchName` thread)

lib/test.js
dominykas: To simplify this - I think initially the `base` param should always be empty (if GH will accept that), and the rest should be addressed as we tackle the existing branches case.

ghinks:
I'll look at this with respect to #93 tomorrow. Today I need to prep or the talk for OpenJSWorld.

lib/test.js
ghinks: Not quite done yet, I'm going for the merging into the default branch main or whatever the default is set to. ( the better DX experience needs another change which I'm going to try for tomorrow.

ghinks: @dominykas I have changed the **createPR** function to go for the easiest option which is merging back into the **default** branch (main more than likely). I want to get this PR in this simpler form and then look at the options you have above.

ghinks: @dominykas , I altered the code to raise the PR against the default branch and added changes to the tests to cover this most simple scenario. I too want to get this PR merged and maybe covering the other edge cases should be handled in another follow up PR?

ghinks:
@dominykas I had some extra time so I have addressed the case where the dependency feature branch name could be pre-existing in the dependent repo, and altered the PR raised to be wiby-feature-x into feature-x in that case.

dominykas:
wiby test

dominykas:
(don't really expect ☝️ to work, but still hoping that maybe GH fixed the issue I was having...)

dominykas (approved):

Issue #46, opened 2020-10-21
Edge case handling


dominykas:
- [ ] `test` should verify the branches do not already exist before pushing
- [ ] `result` should not crash on branches which are not found

PR #94, opened 2021-04-18
Handle a couple of edge cases

dominykas:
This is mostly to get by brain working to figure out what to do when a target branch already exists.

- fixes a `wiby result` crash when a branch already exists (see #46)
- fixes a double-prefix situation when starting point is a branch that's already named `wiby-*` (see #78)


test/fixtures/http/result-command-missing-branch.js
```diff
@@ -0,0 +1,63 @@
+'use strict'
+
+/**
+ * Mocks of HTTP calls for "wiby result" command positive flow
```
rodion-arr, line 4: Maybe it worth to drop few additional words here about this new flow?

rodion-arr (approved):

ghinks:
I pulled it down locally and ran and verified it. Even against node 16. 

The intent to check to see if the branch already exists and then to set the pipeline status enum to **MISSING** is clear.

This is a general comment, not specific to this PR, our tests are getting complicated and at some point will reach the **very difficult to understand** stage. For now not an issue, but my counsel is that if this were to become successful and adopted some re-work may be necessary.

ghinks (approved):
I'm good with these changes, at some point in the future the tests will need re-structuring or documenting so that others can follow in our footsteps.

dominykas:
Agreed on the testing bit. The amount of mocks one has to set up when testing things like this is quite annoying... but testing against a real GH API is also not exactly viable, since you end up with failures due to networking flakes, and it's hard to also ensure a clean slate, and then you also can't run many things in parallel...

So yeah, no easy solution. But we'll come up with something eventually, I'm sure.

PR #100, opened 2021-06-28
fix: correct 404 detection

dominykas:
It seems this started failing some 2 weeks ago due to one of the octokit deps changing


ljharb (approved):
Which dep had a breaking change in a non-major?

rodion-arr (approved):

dominykas:
I didn't dig through it, but it must have been something under octokit.

ghinks (approved):

PR #101, opened 2021-06-28
Re-use existing branches

### LINKED ISSUE #95: Existing branch handling
Scenarios:

1. Suppose you're on a branch `some-branch` on the parent repo
2. You run `wiby test`
3. One of the dependent repos already has a branch called `wiby-some-branch`
	a) It could be that you already ran `wiby test` and did not `wiby clean` afterwards (most common)
	b) It could be that whoever was testing, wanted to make additional changes on the branch to get the tests to pass with the linked parent
	c) It could be that it exists, because multiple dependencies are being tested 

Thinking out loud:

- In cases (b) and (c) we probably need to link the parent into package.json and push a commit into an existing branch; if the parent is already linked, then we probably want to push an empty commit (`git commit -m "This is a blank commit" --allow-empty`) to re-trigger the builds.
- In case (a), we probably want to delete the branch, recreate it from `HEAD` and link the parent again.
	- The problem here is that we don't know if we're dealing with (a) or (b)/(c) - in case of (b)/(c) we want to keep all changes
	- The simple approach would be to not rebase, however that may mean that `wiby test` is running against outdated code (i.e. it no longer answers the question "will I break you"...)
	- Maybe the right answer is to always attempt to update the existing branch with `master`, before pushing a new commit? Can that be done via the API? We could then notify the user about the conflicts, if any, via `wiby results`?

---

Related issues:

- https://github.com/pkgjs/wiby/issues/54
- https://github.com/pkgjs/wiby/issues/78

dominykas:
Closes #95


rodion-arr (approved):
👍

ghinks:
As part of simple due diligence I pulled the branch and ran the tests locally. For me I see issues with test/internals/github-integration.js. 

I am looking further.

Screen Shot 2021-07-19 at 11 00 03 AM
Screen Shot 2021-07-19 at 10 58 17 AM

dominykas:
> For me I see issues with test/internals/github-integration.js.

These tests need a valid `GITHUB_TOKEN` exposed. They get skipped if there is no token at all - maybe you had one for some other GH instance?

dominykas:
@ghinks as these tests are passing for me locally, and also in CI, I'll assume that all is fine and merge this - ping me if you want to dig deeper to see what's going in there?

Issue #96, opened 2021-04-18
Github Workflow: cleanup when parent PR is closed


dominykas:
After a PR is closed/merged, we should automatically run `wiby clean` for that branch.

This implies support for `wiby clean --branch [specific name]` (at the moment, `wiby clean` works off the currently active parent branch), as the branch to be cleaned up will be missing.

Bonus points: support `wiby clean` comment to trigger the cleanup as well.

PR #106, opened 2021-09-27
Use `wiby test` label as a trigger

dominykas:
Implemented as described in https://github.com/pkgjs/wiby/issues/56#issuecomment-887602099

I might have to review the interactions with the checks API - something changed in Github and I was not getting the main `wiby` status displayed on the PR, but otherwise this works.

Some other notes to follow up:

- ~~Does it work OK when PR is opened from a fork? Mainly does the `checkout` action step, which uses the `github.event.pull_request.head.ref`, check out the correct branch/give it the correct name?~~ It does not #107 
- ~~Still need to figure out an option for retest~~ Deferred to https://github.com/pkgjs/wiby/issues/109
- ~~Still need to figure out when is a good time to do `wiby clean`~~ Forgot it's covered in #96 
- ~~Probably need to remove the label if neither test nor results run~~ Covered in https://github.com/pkgjs/wiby/issues/109

cc @zekth

### DIFF
--- a/.github/workflows/wiby.yaml
+++ b/.github/workflows/wiby.yaml
@@ -61,7 +61,6 @@ jobs:
             previews:
               - antiope
           name: ${{ env.WIBY_CHECK_NAME }}
-          details_url: "${{ github.event.repository.html_url }}/actions/runs/${{ github.run_id }}"
           head_sha: ${{ github.event.pull_request.head.sha }}
           status: "queued"
         env:
@@ -101,7 +100,6 @@ jobs:
             previews:
               - antiope
           name: ${{ env.WIBY_CHECK_NAME }}
-          details_url: "${{ github.event.repository.html_url }}/actions/runs/${{ github.run_id }}"
           head_sha: ${{ github.event.pull_request.head.sha }}
           conclusion: "failure"
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/wiby.yaml
```diff
@@ -104,49 +109,44 @@ jobs:
...
               - antiope
           name: ${{ env.WIBY_CHECK_NAME }}
-          details_url: "${{ steps.pr.outputs.repo_url }}/actions/runs/${{ github.run_id }}"
-          head_sha: ${{ steps.pr.outputs.head_sha }}
+          head_sha: ${{ github.event.pull_request.head.sha }}
           status: "in_progress"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Clean up the trigger label
```
zekth, line 122: Why not using the provided action ?
```yaml
      - uses: actions-ecosystem/action-remove-labels@v1
        with:
          labels: wiby
          github_token: ${{ secrets.GITHUB_TOKEN }}
```

dominykas: That's not an official action, is it?

Whereas `octokit` is a much more well known project.

zekth: Oh my bad i thought it was. Let's rely on octokit

zekth (approved):

rodion-arr (approved):

PR #110, opened 2021-11-09
chore: Add Node.js v17 to CI

rodion-arr:


package.json
```diff
@@ -52,6 +52,6 @@
   },
   "support": true,
   "engines": {
-    "node": "^16 || ^14 || ^12"
+    "node": "^17 || ^16 || ^14 || ^12"
```
dominykas, line 55: I think it's OK to add it to the test matrix, but I wouldn't advertise it as supported?

.github/workflows/ci.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [ 16, 14, 12 ]
+        node-version: [ 17, 16, 14, 12 ]
```
dominykas, line 20: Ideally we'd switch this to pkgjs/action, but there's no way to pass in the env vars just yet...

dominykas:
Thanks for the reminder!

package.json
rodion-arr: Agree, reverted

dominykas (approved):

PR #115, opened 2022-01-02
Use pkgjs/action for testing

dominykas:
This needs https://github.com/pkgjs/action/pull/33 to work, so I'll leave this as draft until then.


rodion-arr (approved):
Nice!

PR #140, opened 2024-07-02
security: rely on `sha` rather than `ref`

RedYetiDev:
Relying on a PR's `sha` is more robust than relying on it's `ref`. The `sha` is unique to a commit, while the `ref` is unique to a branch.


ljharb:
The ref resolves to a sha, so it's identically robust.

RedYetiDev:
first off, I didn't mean to close this, whoops!

Secondly, the `ref` resolves the branch name (I.E. `main`), which is not as robust as an `sha` which is constant.

ljharb:
I'm not sure that's valuable here - the sha can be checked after the run.

Issue #147, opened 2025-02-14
Test failing in the latest LTS versions


bjohansebas:
There is a test failing in the latest LTS versions, I will work on fixing it.

![Image](https://github.com/user-attachments/assets/4048d90f-c9d6-46c1-9f24-0da415c33540)

PR #148, opened 2025-02-14
fix(tests): update expected error for invalid JSON config

bjohansebas:
Since the error message has changed when reading a JSON with `require`, the best approach would be to not check what message the error contains.

Closes: #147

### DIFF
--- a/test/config.js
+++ b/test/config.js
@@ -7,8 +7,12 @@ const gitFixture = require('./fixtures/git')
 
 const wiby = require('..')
 
+const semver = require('semver')
+const isGTE20 = semver.satisfies(process.versions.node, '>= 20')
+const isGTE22 = semver.satisfies(process.versions.node, '>= 22')
+
 const invalidConfigs = {
-  'fail-bad-json.json': 'Unexpected end of JSON input',
+  'fail-bad-json.json': isGTE20 ? `${path.resolve(path.join(__dirname, 'fixtures', 'config', 'fail-bad-json.json'))}: Expected ':' after property name in JSON at position 23${isGTE22 ? ' (line 2 column 1)' : ''}"` : 'Unexpected end of JSON input',
   'fail-unknown-root-key.json': '"not-allowed" is not allowed',
   'fail-unknown-dependent-key.json': '"dependents[0].sub-key-not-allowed" is not allowed',
   'fail-dependent-not-url.json': '"dependents[0].repository" must be a valid uri'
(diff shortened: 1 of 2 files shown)


test/config.js
```diff
@@ -8,7 +8,7 @@ const gitFixture = require('./fixtures/git')
 const wiby = require('..')
 
 const invalidConfigs = {
-  'fail-bad-json.json': 'Unexpected end of JSON input',
+  'fail-bad-json.json': null,
```
ljharb, line 11: maybe instead of checking the message, we should check for the error `.code`?

bjohansebas: In this case, .code is undefined, I don't know why, at least in Node 22.14.

bjohansebas: Syntax errors do not have a .code property, so we cannot use such a property

ljharb: fair. maybe then, a regex that covers all of the expected message flavors?

bjohansebas: I wouldn't want to create a regex that covers both the message, Unless we stop supporting node <20, which would be just a single message to verify.

ljharb: why wouldn't it be better to check for one of two known messages, rather than not checking at all?

bjohansebas: I tried using regex for the message in tap, but it seems like it's not working, though I think the solution is better now.

test/config.js
```diff
@@ -7,8 +7,10 @@ const gitFixture = require('./fixtures/git')
 
 const wiby = require('..')
 
+const nodeVersion = Number(process.versions.node.split('.')[0])
```
ljharb, line 10:
```suggestion
const semver = require('semver');
const isGTE20 = semver.satisfies(process.version, '>= 20');
```
to be more reliable?

bjohansebas: It would mean installing another dependency for something so simple, and since it's only used in this part, I don't think adding that dependency is justified.

ljharb: There’s zero cost to adding extra dev deps, and i think doing a guaranteed-correct check seems worth it ¯\_(ツ)_/¯

bjohansebas: done

ljharb (approved):

bjohansebas:
God, the message changed a bit from version 20 to version 22.

ljharb:
k, i rebased this, fixed the tests for node 20/22, and pinned a dev dep to hopefully fix node 12

PR #149, opened 2025-02-15
docs: add usage instructions for `wiby close-pr` command

bjohansebas:
Run `npm run generate-docs`, maybe we could create a workflow to open a PR every time the options are updated.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -22,6 +22,25 @@ Options:
 ```
 
 
+## `wiby close-pr`
+
+
+
+Use this command to close the PRs raised against your dependents. wiby will go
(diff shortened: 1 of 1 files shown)


ljharb (approved):
seems reasonable

bjohansebas:
this package doesn't even work for Node 12 anymore, maybe we should stop supporting those versions of Node.js?

And the errors in the CI for Node LTS should be fixed with #148.

ljharb:
I think it's a good idea to get the tests passing on all supported node lines before dropping support for any of them.

bjohansebas:
Yep, although well, this is not the responsibility of this PR

PR #150, opened 2025-02-15
fix: fix support node 12

bjohansebas:
After simple-git@3.24.0, they import the modules with `node:`, which breaks compatibility with Node 12.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "dotenv": "^16.0.0",
     "git-url-parse": "^11.1.2",
     "joi": "^17.2.1",
-    "simple-git": "^3.0.0",
+    "simple-git": "3.23.0",
     "yargs": "^17.0.0"
   },
   "devDependencies": {


package.json
```diff
@@ -36,7 +36,7 @@
     "dotenv": "^16.0.0",
     "git-url-parse": "^11.1.2",
     "joi": "^17.2.1",
-    "simple-git": "^3.0.0",
+    "simple-git": "3.23.0",
```
ljharb, line 39:
```suggestion
    "simple-git": "=3.23.0",
```

ljharb (approved):

ljharb:
Has an issue been filed with simple-git for a breaking change in a minor version? (omitting `engines` is the same as `*`, meaning, all node versions are supported)

bjohansebas:
@ljharb Apparently, they only test with LTS versions.

ref: 
https://github.com/steveukx/git-js/issues/997
https://github.com/steveukx/git-js/blob/95c556f3c13cb21061025bdb37c3af197a28c0e9/.github/workflows/ci.yml#L17

ljharb (approved):

bjohansebas:
This can be merged now, right?

ljharb:
Is there an open PR to fix tests on 20 and 22 i could land first?

bjohansebas:
That's #148, although it seems like it only fixes it for Node version 22.

ljharb:
turns out it's v3.11, and i snuck that in in 5631d73, so i think this can be closed (sorry about that)

Issue #49, opened 2020-10-27
Provide a commit sha to be tested


dominykas:
Ref: https://github.com/pkgjs/wiby/blob/master/docs/DRAFT-FLOW.md step 3

Support `wiby test --commit=[sha]` (if left out - keep current behavior and test currently checked out `HEAD`)

PR #151, opened 2025-02-17
feat: add `sha` option for testing and update schema to include branch

bjohansebas:
The option to choose which branch to use as a reference for testing is added, as the main branch will not always be the one selected.

Closes https://github.com/pkgjs/wiby/issues/49

### DIFF
--- a/bin/commands/result.js
+++ b/bin/commands/result.js
@@ -10,6 +10,10 @@ exports.builder = (yargs) => yargs
     type: 'string',
     conflicts: 'config'
   })
+  .option('sha', {
+    desc: 'Commit or branch that was chosen for testing',
+    type: 'string'
+  })
   .option('config', {
     desc: 'Path to the configuration file. By default it will try to load the configuration from the first file it finds in the current working directory: `.wiby.json`, `.wiby.js`',
     type: 'string'
@@ -18,7 +22,7 @@ exports.builder = (yargs) => yargs
 exports.handler = async (params) => {
   const config = params.dependent
     ? {
-        dependents: [{ repository: params.dependent }]
+        dependents: [{ repository: params.dependent, sha: params.sha || 'HEAD' }]
       }
     : wiby.validate({ config: params.config })
 
--- a/bin/commands/test.js
+++ b/bin/commands/test.js
@@ -16,6 +16,10 @@ exports.builder = (yargs) => yargs
     type: 'boolean',
     conflicts: 'config'
   })
+  .option('sha', {
+    desc: 'Test against a specific commit or branch',
+    type: 'string'
+  })
   .option('config', {
     desc: 'Path to the configuration file. By default it will try to load the configuration from the first file it finds in the current working directory: `.wiby.json`, `.wiby.js`',
     type: 'string'
@@ -24,7 +28,7 @@ exports.builder = (yargs) => yargs
 exports.handler = (params) => {
   const config = params.dependent
     ? {
-        dependents: [{ repository: params.dependent, pullRequest: !!params['pull-request'] }]
+        dependents: [{ repository: params.dependent, pullRequest: !!params['pull-request'], sha: params.sha || 'HEAD' }]
       }
     : wiby.validate({ config: params.config })
 
--- a/lib/github.js
+++ b/lib/github.js
@@ -15,17 +15,27 @@ const graphqlWithAuth = graphql.defaults({
   }
 })
 
-module.exports.getPackageJson = async function getPackageJson (owner, repo) {
+module.exports.getPackageJson = async function getPackageJson (owner, repo, sha) {
   try {
     const resp = await graphqlWithAuth({
-      query: queries.getPackageJson,
-      owner: owner,
-      repo: repo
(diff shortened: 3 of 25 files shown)


bin/commands/result.js
```diff
@@ -10,6 +10,11 @@ exports.builder = (yargs) => yargs
     type: 'string',
     conflicts: 'config'
   })
+  .option('branch', {
+    desc: 'Branch that was chosen for testing',
+    type: 'string',
+    conflicts: 'config'
```
ljharb, line 16: why would this conflict with config? Id expect the config to be overridden by command line flags

bjohansebas: I was just following how the other options were programmed, although what you're saying makes sense.

bin/commands/test.js
```diff
@@ -24,7 +28,7 @@ exports.builder = (yargs) => yargs
 exports.handler = (params) => {
   const config = params.dependent
     ? {
-        dependents: [{ repository: params.dependent, pullRequest: !!params['pull-request'] }]
+        dependents: [{ repository: params.dependent, pullRequest: !!params['pull-request'], branch: params["branch"] || "HEAD" }]
```
ljharb, line 31: nit:
```suggestion
        dependents: [{ repository: params.dependent, pullRequest: !!params['pull-request'], branch: params.branch || "HEAD" }]
```

bin/commands/result.js
```diff
@@ -18,7 +22,7 @@ exports.builder = (yargs) => yargs
 exports.handler = async (params) => {
   const config = params.dependent
     ? {
-        dependents: [{ repository: params.dependent }]
+        dependents: [{ repository: params.dependent, branch: params["branch"] || "HEAD" }]
```
ljharb, line 25:
```suggestion
        dependents: [{ repository: params.dependent, branch: params.branch || 'HEAD' }]
```

ljharb (approved):
seems good overall

bin/commands/result.js
bjohansebas:
```suggestion
        dependents: [{ repository: params.dependent, branch: params?.branch || 'HEAD' }]
```

ljharb: same here re `.dependent`

bin/commands/test.js
bjohansebas: In this case, the parameters can be nullish
image

ljharb: then you need `?.[` on the pull-request lookups also

bin/commands/result.js
```diff
@@ -18,7 +22,7 @@ exports.builder = (yargs) => yargs
 exports.handler = async (params) => {
   const config = params.dependent
     ? {
-        dependents: [{ repository: params.dependent }]
+        dependents: [{ repository: params.dependent, branch: params?.branch || 'HEAD' }]
```
ljharb, line 25: either params.dependent needs optional chaining, or params.branch doesn't

bin/commands/test.js
```diff
@@ -24,7 +28,7 @@ exports.builder = (yargs) => yargs
 exports.handler = (params) => {
   const config = params.dependent
     ? {
-        dependents: [{ repository: params.dependent, pullRequest: !!params['pull-request'] }]
+        dependents: [{ repository: params.dependent, pullRequest: !!params['pull-request'], branch: params?.branch || "HEAD" }]
```
ljharb, line 31: either params.dependent and params['pull-request'] both need optional chaining, or params.branch doesn't

ljharb (changes requested):

bjohansebas:
Optional chaining is not available for Node 12 :(

ljharb:
but the pre-existing code also suggests it's not needed. maybe it's just IDE types that are wrong?

bjohansebas:
Probably

bjohansebas:
I don't understand why it doesn't work for Node 12 now, there hasn't been any change that should break compatibility.

ljharb:
try rebasing instead of merging and force push?

bjohansebas:
I've found the error :)

lib/github.js
```diff
@@ -51,10 +61,11 @@ module.exports.getDefaultBranch = async function (owner, repo) {
   return resp.repository.defaultBranchRef.name
 }
 
-module.exports.getShas = async function getShas (owner, repo) {
+module.exports.getShas = async function getShas (owner, repo, branch) {
   const resp = await octokit.repos.listCommits({
     owner,
     repo,
+    sha: branch,
```
ljharb, line 68: maybe this argument should be called "sha" or "commitish" to indicate it doesn't have to be a branch name?

bjohansebas: SGTM

bjohansebas:
I just realized that this is related to #49, although I think 'sha' is a better name than 'commit'.

bjohansebas:
@ljharb Can you take another look, and maybe merge it if it seems good to you?

bjohansebas:
Can we merge this, please?

ljharb (approved):

PR #154, opened 2025-02-20
[WIP]: feat: add mode option for testing

bjohansebas:
The idea is to make this tool work without a GitHub token, allowing us to test those tools we don't have access to, but don't want to unintentionally break.


bjohansebas:
I think the best way to try doing this is to generate a tarball of the local package, use that file as a dependency, and run tests.

bjohansebas:
I've been thinking about what the workflow could be like, maybe something like this could work:

image

PR #158, opened 2025-03-23
feat: add linting job to CI workflow and simplify test script

bjohansebas:
The linter is not the tests. I want to run the tests peacefully without worrying about the linter. I can worry about the linter later after finishing my changes.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,5 +27,7 @@ jobs:
 
   test:
     uses: pkgjs/action/.github/workflows/node-test.yaml@main
+    with:
+      upgrade-policy: all
     secrets:
       test-secrets: '{"GITHUB_TOKEN":"${{ secrets.GITHUB_TOKEN }}"}'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,6 +9,21 @@ on:
   pull_request:
 
 jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+
+      - name: Install dependencies
+        run: npm install --ignore-scripts --include=dev
+
+      - name: Run lint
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -13,9 +13,7 @@
     "action-wiby-test": "npm install --production && ./bin/wiby test",
     "action-wiby-result": "npm install --production && ./bin/wiby result",
     "lint": "standard",
-    "pretest": "[ \"$NODE_LTS_LATEST\" != \"\" ] && [ \"$MATRIX_NODE_VERSION\" != \"$NODE_LTS_LATEST\" ] && echo 'Skipping linting' || npm run lint",
-    "test": "npm run tests-only",
-    "tests-only": "tap",
+    "test": "tap",
```
ljharb, line 16: `npm test` should always run everything. you run `npm run tests-only` if you only want to run the tests.

.github/workflows/ci.yaml
```diff
@@ -9,8 +9,25 @@ on:
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+
+      - name: Install dependencies
+        run: npm install --ignore-scripts --include=dev
+
+      - name: Run lint
+        run: npm run lint
```
ljharb, line 26: i believe we already run the linter via `pretest`

ljharb (changes requested):

.github/workflows/ci.yaml
bjohansebas: So, should the linter run after the tests using `posttest`? I've never agreed that the linter should run with the tests—it should be something separate and in a different job.

ljharb: Locally, *everything* should run with `npm test`.

The CI jobs should be configured so that the linter runs once, and the tests run on every supported node version (minor or major).

Thus, `npm test` runs everything, and CI workflows run more granular scripts.

bjohansebas: I don't think the linter should live under the test command.

ljharb: Why not? That's how every project I've maintained or contributed to works.

PR #159, opened 2025-03-23
feat: include files in package.json for better distribution

bjohansebas:
We are sending extra files, ideally, we should only send what is necessary to npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,12 @@
   "directories": {
     "test": "test"
   },
+  "files": [
+    "lib",
+    "bin",
+    ".github/workflows/wiby.yaml",
+    "package-support.json"
+  ],
   "scripts": {
     "action-wiby-test": "npm install --production && ./bin/wiby test",
     "action-wiby-result": "npm install --production && ./bin/wiby result",


package.json
```diff
@@ -9,6 +9,12 @@
   "directories": {
     "test": "test"
   },
+  "files": [
+    "lib",
+    "bin",
+    ".github/workflows/wiby.yaml",
+    "package-support.json"
+  ],
```
ljharb, line 17: 1. the `files` field is dangerous and should never be used; `npmignore` is how files are kept out of the published package
2. tests should be published in a package so that `npm explore foo && npm install && npm test` always works
3. github workflows never need to be published.

ljharb (changes requested):

package.json
bjohansebas: > the files field is dangerous and should never be used; npmignore is how files are kept out of the published package

Sure, I also agree with that, but I see that many packages use `files` instead of `.npmignore`

> tests should be published in a package so that npm explore foo && npm install && npm test always works

That's new to me, and it's the first time i've heard it. So, let's send the tests.

> github workflows never need to be published.

with Wiby's current design, we need to send that file. I want to change the way Wiby runs in GitHub Actions, but that will take time.

ljharb: I'm confused why the published wiby tarball needs to contain a github action, since actions are only run from repos?

bjohansebas: By a command we used that copied that same GitHub Action, **wiby github-workflow install**

https://github.com/pkgjs/wiby/blob/main/USAGE.md#wiby-github-workflow-install

ljharb: ah ok, good catch :-)

PR #160, opened 2025-03-24
Release: 0.10.2

bjohansebas:
Maybe it's time to make a release with what we have. There are several bug fixes, support for Node 12 is fixed, and support is expanded to Node >=18 and a new feature is added (https://github.com/pkgjs/wiby/pull/151).  

@ljharb, could you do the release, please?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wiby",
-  "version": "0.10.1",
+  "version": "0.10.2",
   "description": "Will I Break You?",
   "bin": {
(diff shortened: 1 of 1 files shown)


ljharb:
If there's no breaking changes, it should be 0.10.2?

ljharb:
Yes, but v0 versions are special. in vX.Y.Z, v0.X.Y, and v0.0.X, X is the major, Y minor, Z patch.

ljharb (approved):

PR #161, opened 2025-03-24
Ensure GitHub token

bjohansebas:
To prevent commands from being executed if there is no GitHub token in the environment variables.


ljharb (approved):

bjohansebas:
Can we merge this, please?

PR #162, opened 2025-05-12
move simple-git from devDependencies to dependencies in package.json

bjohansebas:
This is a dependency since it's not used for testing, but for essential package functionality.


bjohansebas:
The CI is failing because the token no longer works, see https://github.com/nodejs/package-maintenance/issues/629. 

The previous error that was showing was also  (https://github.com/pkgjs/wiby/actions/runs/14960947688/job/42022736967?pr=144)
imagen

